PR #6, opened 2018-10-08
Fix for some typos

mathias-arm:


gmandyam:
@mathias-arm 

Thanks for providing this.  It is now merged.

PR #7, opened 2019-04-16
Added Claims Characteristics section

laurencelundblade:


draft-ietf-mandyam-eat.md
```diff
@@ -260,6 +260,111 @@ signing. In this case the manufacturer should perform the verification,
...
+to any operating system. They should be applicable to
+multiple large high-level operating systems from different
+vendors. They should also be applicable to multiple 
+embedded operating systems from multiple vendors and
+everything in between.
+
+Claims should not be defined such that they are specific to any SW
+environment or programming language.
+
+Claims should not be defined such they are not specific to any chip or
```
setrofim:
```suggestion
Claims should not be defined such they are specific to a chip or
```
or
```suggestion
Claims should be defined such they are not specific to any chip or
```

draft-ietf-mandyam-eat.md
```diff
@@ -260,6 +260,111 @@ signing. In this case the manufacturer should perform the verification,
...
+security use cases just the same. For example, an app in a mobile app
+download system may generate tokens.
+
+The claims included in this document are on the basis of whether they
+are useful and interoperable.  No claims have been excluded because
+they are not useful on high-security environments and entities.
+
+### Reuse of Extant Data Formats
+
+Where possible, claims should already standardized data items,
```
SimonFrost-Arm: Typo:  claims should +use+ already standardized

draft-ietf-mandyam-eat.md
```diff
@@ -260,6 +260,111 @@ signing. In this case the manufacturer should perform the verification,
...
+
+In some rare cases, it may be better to use the encoding as is. For
+example, signed X.509 certificates and CRLS should be carried as is
+and not re encoded.
+
+### Proprietary Claims
+
+General compliance with EAT allows arbitrary proprietary claims 
+to be added. In many cases proprietary claims will be the
+easiest and most obvious way to proceed.
```
SimonFrost-Arm: Add comment to the effect that proprietary claims should not replace or overlap standard claims unless the implementation is significantly different?

henkbirkholz:
I would recommend to remove the "Some examples of profiles that may come to exist:" content. A specification or even proposed standard is not a good place for listing "work that may or may not be done", I think. Other SDO actually forbid this type of content rather strictly. I am not sure what the exact IETF rules are, tbh.

draft-ietf-mandyam-eat.md
nedmsmith: Shouldn't the language be "...such THAT they..."?

Not sure why the EAT draft should give guidance on what belongs as a claim (or not). Doesn't the existence of the claim in the spec resolve whether or not it should exist?

nedmsmith:
I assert that existence in a specification justifies its existence!

laurencelundblade:
I take Ned's point that it is a bit circular to justify the existence of claims that have been agreed upon in the standard. However I think the text is important and helpful and should be in the document somewhere.

One possibility is the section for IANA as advice for claims registration. This is still odd though since there is no EAT IANA registry as we re use CWT.

laurencelundblade:
@henkbirkholz on the examples to be built, I agree that this is probably not the right way to frame it. What about just listing the as examples with out the "may come to exist" part?

Issue #21, opened 2019-07-21
RAND UEID's - minimum requirements to prevent clashes from different manufacturers


gmandyam:
Current text (-01) does not state assumptions on rand generation that would be necessary to minimize the probability of collisions from different manufacturers.  This could be added to the Security Considerations section.

nedmsmith:
An attacker might purposefully reuse an existing EUID (random or not). The security of attestation should not depend on a convention for whether the generation algorithm includes entropy. Even if values are guessable, there should be an understanding by the Verifier for how to detect that the value is authentic / spoofed. The entity (vendor) creating the EUID is not motivated to issue duplicates. Whether they select a method based on entropy or assigned values should be a vendor specific choice (IMO).

laurencelundblade:
@nedmsmith Addressing your comment from July 22, 2019. The relying party has to trust the UEID based on the fact that it was signed by the attester. They trust the attester because of endorsements. There's nothing more that can be done.

The vendor can choose the type they want, random, MAC address.... They can even switch from one day to the next.

laurencelundblade:
I believe this is fixed with #35

Issue #12, opened 2019-07-03
Should minimum size of UEID be larger than 128 bits?


laurencelundblade:
Ben commented that 128 bits is not enough. Is my analysis correct?

128 bits (16 bytes) is enough for one trillion humans to each have one trillion devices and the probability of collision less than a trillion (log base 10 of 2^ 128 = 38; log base 10 of a trillion is 12; 3 * 12 < 38) So 128 bits seems like enough, but take it out to 256 bits because SHA-256 is popular and good and there is some extra margin.

nedmsmith:
Simply anticipating a namespace large enough to create a universally unique value doesn't ensure there will not be collisions. Often there is a namespace authority that applies a formula for testing uniqueness / collisions. This implies the namespace authority plays a role. For example, X.509 serial numbers are unique within the context of the issuing CA. The issuing CA is the namespace authority. The EUID definition implies namespace authorities exist for example the first 8 bytes of an IMEI identifies the mobile equipment vendor. Presumably the serial numbers are chosen by the vendor. 
It isn't clear that all EUIDs will have bits reserved for namespace authorities and that there will be some way to register them.

laurencelundblade:
PR #35 includes a full probabilistic analysis rather than the poor one in my comment above.

@nedmsmith -- the draft proposes several types of UEIDs, some based on existing on device ID schemes guaranteed by a central database and one based purely on cryptographic quality large random numbers.

laurencelundblade:
Fixed in #35

PR #35, opened 2019-10-07
Up UEID to 256 bits and add design rational appendix

laurencelundblade:
Comments on the analysis are very much requested. 

See issue #12 and #21


nedmsmith:
256-bits is plenty. 128-bits as a minimum. This would allow the use of a digested value. If the namespace isn't digested, then you are relying on the namespace authority to manage the population in some way. If the UEID obfuscates the namespace authority then you have to assume there isn't a namespace authority, which implies using digesting. Hence 128-bit is currently practice for minimum acceptable hash lengths.

richsalz:
This is ridiculously large.  IPv6 uses an identifier space that is half the number of bits

draft-ietf-rats-eat.md
```diff
@@ -1019,12 +1023,139 @@ is shown.
...
+The trillion record database size models an easy-to-imagine reality
+over the next decades. The quadrillion record database is a worst case
+that might not come to pass, but would seem to be possible and thus
+should be accommodated. The highly speculative 100 quadrillion case
+perhaps involve nanorobots for every person and livestock animal and
+domesticated bird. It is included to round out the analysis.
+
+Note that the items counted here certainly do not have IP address and
+are not individually connected to the network. They may be connected
+to internal buses, via serial links, blue tooth and so on.  This is
```
thomas-fossati: typo
```suggestion
to internal buses, via serial links, Bluetooth and so on.  This is
```

draft-ietf-rats-eat.md
```diff
@@ -1019,12 +1023,139 @@ is shown.
...
+
+The proper formula for the collision calculation is
+
+       p = 1 - e^{-k^2/(2n)}
+    
+       p   Collision Probability
+       n   Total possible population
+       k   Actual population
+
+However, for the very large values involve here, this formula requires floating
```
thomas-fossati:
```suggestion
However, for the very large values involved here, this formula requires floating
```

draft-ietf-rats-eat.md
```diff
@@ -1019,12 +1023,139 @@ is shown.
...
+point precession higher than commonly available in calculators and SW so this
```
thomas-fossati:
```suggestion
point precision higher than commonly available in calculators and SW so this
```

draft-ietf-rats-eat.md
```diff
@@ -1019,12 +1023,139 @@ is shown.
...
+Clearly 128 bits is not enough for the quadrillion-record database. Clearly
+192 bits is enough for every case.
+
+Note that even the rate of change of the database were much lower with one
+thousandth of it changing per year, that would not improve the rate for
+the one quadrillion record database to even one collision per day.  
+
+The choice is for 256-bits to be prepared even for what can’t be
+envisioned today. The table shows 192 bits is enough, but 256 bits,
+32-bytes, is selected because it the output of the common SHA-256 hash
```
thomas-fossati:
```suggestion
32-bytes, is selected because it is the output of the common SHA-256 hash
```

draft-ietf-rats-eat.md
```diff
@@ -1019,12 +1023,139 @@ is shown.
 }
 ~~~~
 
+# UEID Design Rationale
+
+## Collision Probabilitty
```
setrofim: Typo
```suggestion
## Collision Probability
```

laurencelundblade:
I don't think 128 bits being enough for IP addresses implies it is enough for UEIDs. IP addresses get reused serially over time and by NAT. Neither reuse applies to UEIDs. Also, a device with a single IP address may have many sub parts that each have a UEID.  They are connected to each other via a bus, a serial link, BlueTooth or such, not IP.

On the other hand, I don't think the non-correlation with IP address automatically makes UEIDs 256 bits. 256 bits is based on the probabilistic analysis in the appendix.

laurencelundblade:
@nedmsmith This discussion is for RAND type UEIDs as you deduced in your comment above. Appendix B should explicitly say that.

For crypto, 128 bits is usually considered enough today, but there is always an assumption built into designs that it won't be. Also, there is a lot of use of
256 bits today. For example COSE ECDSA signature are minimum SHA-256.

It does seem like UEIDs could start out at 128 bits and transition to 256 bits in ten, twenty or fifty years. All we have to do here and now is say receivers must be prepared for 256 bits. When the time comes a new RFC is issued say 256 bits must be used. The 128-bit and 256-bit UEIDs will coexist just fine. 

I firmly believe we must say now the receivers must be prepared for 256 bits.

The question is probably between these two:

1. Require 256 bits now; waste space for a few decades; no transition will ever be needed.
2. Require only 128 bits now; don't waste space; have a transition if and when the need arises

richsalz:
Waste space for a few decades?  The IETF is barely 30 years old.

Also, "128 bits is enough" -- don't confuse the DIGEST with the crypto KEYLENGTH.  Few people think 128-bit digest is good enough, but most think 128bit AES, for example, is good enough.

draft-ietf-rats-eat.md
laurencelundblade: Grammar

laurencelundblade:
@richsalz 

Are you in favor of 1 or 2? I assume 1, but don't understand your comment about IETF being 30 years old.

Agree that my discussion of bits used by crypto today is not very useful here. The only really useful point is that we build in agility to protocols today so they can use more bits when needed.

richsalz:
The IETF is 30 years old, planning for "waste space for a few decades" when the whole thing has only been around for three seems, well, optimistic? Naive? Wrong? Silly? Because of this, I think 1 is the wrong trade-off and I support 2. (Recall the Internet has already changed core protocols multiple times)

nedmsmith:
For me the UEID discussion is strange because UEIDs seem to overload identity and integrity. An IMEI number encodes some information attributes that provides uniqueness within telco networks. There is not expectation of integrity that is provided by some other structure. 
In a perfect world, the structure would be defined using a DDL like CDDL and everything consuming it would support variable length structures. However, the reality is systems are built with fixed length assumptions and so we see schemes that encode multiple attributes into a fixed length structure. If there are many attributes needed to make something "unique" then some favor using a hash algorithm since it can reduce anything into a fixed size statistically unique value. The decision to go with 128 or 256 bits has more to do with the intended scope of identifier use. Since UEID intends to combine lots of different identifiers (each being unique within their respective scope) the union of all them creates a larger scope context for understanding uniqueness (aka resistance to collision). Since, nobody (AFAIK) has done an analysis to determine the actual size of the union-of-all-EIDs (UEID), we don't know if 128-bits still has the collision resistance properties it needs. 256 seems safe at least for now. 

Alternatively, EAT / RATS could allow claims that more directly define the identifier type and context. For example, there could be an IMEI claim. 

I'm not sure the intended value of UEID actually exists.

laurencelundblade:
UEIDs can be IMEIs (read the document). UEIDs are only about identity, never about integrity.

Issue #22, opened 2019-07-21
Sec. 3.7.6 (-01):  Should debug disable states be an array or a choice option?


gmandyam:
Sec. 2.2.2 of RFC 8610 allows for a choice expression based on a '/' delimiter between options.  In order to avoid expression of potentially conflicting debug disable states (e.g. debug full permanent disable being true while debug disabled since boot being false), this could be an alternative.

laurencelundblade:
I think I agree with this with the concept being that you claim the highest level of debug disable that is true for the device at the moment. That means selecting a higher level of debug disabling requires all the lower levels of debug disabling are true.

laurencelundblade:
Fix with #36

PR #36, opened 2019-10-11
Revamp debug disable so it is levels, not Booleans

laurencelundblade:
This addresses #22


SimonFrost-Arm:
The change to a set of levels is a good one as it solves inconsistencies with the previous multi claim set.

PSA calls out a further concept to be tracked for debug: how intrusive the level of debug is, ie. whether it just applies to the non-secure application state or to a higher level of security. The implication of the latter is that the configuration of the device/entity could be compromised making it unsuitable for further trust.

draft-ietf-rats-eat.md
```diff
@@ -555,56 +555,86 @@ claim made here is solely a self-claim made by the Entity Originator.
...
+If any debug facility, even manufacturer hardware diagnostics, is
+currently enabled, then this level must be indicated.
 
-### Debug Disabled Since Boot
+#### Disabled
 
-This claim indicates whether debug capabilities for the entity were
-not disabled in any way since boot (i.e. value of 'true').
+This level indicates all debug facilities are currently disabled. It
+may be possible to enabled them, but they are currently disabled.
```
thomas-fossati:
```suggestion
may be possible to enable them in the future, and it may also be possible that they were enabled in the past after the device booted, but they are currently disabled.
```

laurencelundblade:
@SimonFrost-Arm 

I am scared to try to express debug intrusiveness in claims because I'm not sure it can be easily done in a general way. Different CPU and system architectures might be too different. Are you suggesting we try? 

To go into more detail... Some HW I know of has a whole facility for HW diagnostic and verification. It just helps the HW manufacturer know that the chip actually works. It might be used during RMA to figure what is wrong.

Another is a facility like ARM CoreSight. I believe that is connected to through JTAG.

When you talk about PSA secure vs non-secure are you talking about HW or SW debug facilities?

henkbirkholz:
I do not understand how this is supposed to work, tbh. Is this supposed to be a choice of integers that represent a range? If so, it is not doing that. If not, I do not understand what is attempted.

>     debug_disable_level = (
>         not_disabled: 0, 
>         disabled: 1,
>         disabled_since_boot: 2,
>         permanent_disable: 3,
>         full_permanent_disable: 4
>     )

Branch cddl-fixes includes a "fixed" version that still uses bool. If I understand the intent here, I can incorporate that in the cddl-fixes branch.

SimonFrost-Arm:
The intent of the change is to remove the possibility of inconsistency that could occur with the multi bool approach e.g. it would be possible to have a token containing both debug_disabled_since_boot and debug_permanently disabled - which is the case? Replacing the multiple values with a single multi-level value makes the declaration much clearer.

henkbirkholz:
If all possible options are mutual exclusive it would look like this:

> debug_disable_level /= not_disabled
> debug_disable_level /= disabled
> debug_disable_level /= disabled_since_boot
> debug_disable_level /= permanent_disable
> debug_disable_level /= full_permanent_disable
> 
> ;this is the CBOR specific value list
> not_disabled = 0
> disabled = 1
> disabled_since_boot =2
> permanent_disable = 3
> full_permanent_disable = 4

henkbirkholz:
After I understood the intent via PR #43 and if you really want to use the "&" prefix, the corresponding CDDL for this PR should probably look like this:

> boot_state_claim = ( boot_state => boot_state_type )
> 
> debug_disable_level = (
>   not_disabled,
>   disabled,
>   disabled_since_boot,
>   permanent_disable,
>   full_permanent_disable,
> )
> 
> boot_state_type = [
>   secure_boot_enabled: bool,
>   debug_state: &debug_disable_level,
> ]
> 
> ;this is the CBOR specific value list
> not_disabled = 0
> disabled = 1
> disabled_since_boot = 2
> permanent_disable = 3
> full_permanent_disable = 4

laurencelundblade:
@henkbirkholz 

> debug_disable_level = (
>     not_disabled: 0, 
>     disabled: 1,
>     disabled_since_boot: 2,
>     permanent_disable: 3,
>     full_permanent_disable: 4
> )

This seems OK to me because it is the same as section 2.2.2.2 in RFC 8610.

Seems simpler to do this than the other ways you've proposed. I assume it would work OK for JSON.

I've fixed the other CDDL that was clearly wrong with a commit.

Thanks for the input.

adrianlshaw:
> When you talk about PSA secure vs non-secure are you talking about HW or SW debug facilities?

@laurencelundblade in the PSA definition it can be either or a combination, depending on the hardware platform capabilities. Nothing platform specific is assumed. In fact, "Secure and non-secure" isn't actually the term used. The actual states are more about the debug property relative to the root of trust. See the diagram below, but replace "PRoT" with your favourite root of trust e.g. TEE. 

image

The Decommissioned state is RMA. The "Terminate" process is the point of no return because the root of trust is compromised. 

"Recoverable PRoT debug" is an optional state that covers a case where you have special hardware that hides the root secrets while in that mode, making it possible to re-establish trust (go back to "Secured") once leaving that mode.

laurencelundblade:
@SimonFrost-Arm Good comment. Updated with a new Not Reported debug level that is also the default.

Issue #37, opened 2019-10-16
Sort out cti, jti and nonce claims from CWT and JWT


laurencelundblade:
Which CWT / JWT field is really the nonce?

laurencelundblade:
Addressed in #42

PR #28, opened 2019-08-12
Security Considerations

gmandyam:
First cut

PR #42, opened 2019-10-23
Add nonce claim, update jti/cti text

laurencelundblade:
Addresses #37


draft-ietf-rats-eat.md
```diff
@@ -367,23 +367,42 @@ aggregation and type definition feature of CDDL). In the data model,
...
 date-of-creation of the token.
 
+
+## Nonce Claim (nonce)
+
+All EATs should have a nonce to prevent replay attacks. The nonce is
+generated by the relying party, the end consumer of the token. It is
+conveyed to the entity over whatever transport is in use before the
+token is generated and then included in the token as the nonce claim.
+Note that intrinsically by the nature of a nonce no security is needed
```
thomas-fossati: It is not clear what this means to me.  Besides, in the light of the text proposed in #28
```
[...] Since an EAT may be created by an entity that may not support the same type of transport security as the consumer of the EAT, intermediaries may be required to bridge communications between the entity and consumer.  As a result, it is RECOMMENDED that both the consumer create a nonce, and the entity leverage the nonce along with COSE mechanisms for encryption and signing to create the EAT
```
this should probably be dropped?

laurencelundblade: Hi Thomas,

Have read this over a few times, but I don't know what is not clear to you and what you think should be dropped.

thomas-fossati: (Apologies for the late response -- for some reasons I didn't get the nudge from
GitHub on your reply).   The thing that I don't understand is `Note that
intrinsically by the nature of a nonce no security is needed`.  In fact, I think
it's quite the opposite: the nonce is core to the security of the attestation?
My suggestion is to simply drop the `Note [...] no security is needed`.

laurencelundblade:
@gmandyam in #28 discusses multiple relying parties (RPs) and nonces. There are three ways to solve this. 

1. Each RP can generate their own nonce and send it to the attester. The attester puts them all in. In that case the nonce should be either a bstr with a single nonce or an array of bstr to carry multiple nonces. This is incompatible with the JWT nonce and we'd have to create a new claim it. Call it the multi-nonce claim.

2. The RP's are expected to coordinate around one nonce and the nonce claim is just a single as it is. This is not too difficult. One of them has to generate it and tell the others what it is. This is compatible with the JWT nonce and with the current text.

3. We can define two claim types. One is the nonce as is in JWT and as in this PR. The other is the multi-nonce.

I'm in favor of 2 now. If we really need multiple nonces sometime in the future we can turn it into 3. I guessing multiple nonces will not be that necessary. I can't say that I've seen it any other protocol.

PR #43, opened 2019-11-01
revamp of submods and nested tokens

laurencelundblade:
This makes nested tokens be part of the submods part of the token with a more unified results.

It gets rid of the submod name claim and uses the map label instead.

It adds the attachment claim which describes how securely submod is attached.


draft-ietf-rats-eat.md
```diff
@@ -685,33 +714,81 @@ more security-oriented subsystems like a TEE or a Secure Element.
...
+token.  The subordinate modules must explicitly include all of their
+claims. This is the case even for claims like the nonce and age.
+
+This rule is in place for simplicity. It avoids complex inheritance
+rules that might vary from one type of claim to another. (TODO: fix
+the boot claim which does have inheritance as currently described).
+
+### Security Levels and Attachment
+
+The security level of the non-token subordinate modules should always
```
SimonFrost-Arm: If submods are used structurally within an EAT, they should be able to have the same level of security as the EAT (rather than making lesser security mandatory). May just be unfortunate usage of the word 'should' and can be stated as 'not higher'?

draft-ietf-rats-eat.md
```diff
@@ -685,33 +714,81 @@ more security-oriented subsystems like a TEE or a Secure Element.
...
+of the device.
+
+The opposite may be true for the nested tokens. They usually have
+their own more secure key material. An example of this is an embedded
+secure element.
+
+The submods attachment claim describes how strong the connection is of
+a submodule to the containing module. For stronger attachment types
+the submod might choose to not include claims like the nonce for size,
+but this is not recommended.
```
SimonFrost-Arm: this seems to contradict line 755 which mandates the inclusion of all such claims.
Should there be an explicit statement (or claim value) for whether submod is expected to inherit or not rather than making it implicit on attachment type?

draft-ietf-rats-eat.md
```diff
@@ -656,26 +656,55 @@ seconds that have elapsed since the entity or submod was last booted.
...
+
+    attachment_type = (
+    unspecified: 1,
+    device_internal: 2,
+    pcb_internal: 3,
+    chip_internal: 4
+    )
+    
+    attachment_type_claim = (
+    attachment: attachment_type )
```
SimonFrost-Arm: mismatch here with naming @ line 660

thomas-fossati:
I'm sorry but what we want to achieve with this change is not clear to me. (I've looked for a related Issue to no avail.) Could you please describe the rationale of this change in an Issue (or point me at an existing one, if I overlooked it)?
As a meta-comment I think in general it's good practice to open an Issue first where the problem is explained (maybe just pointing to the corresponding mailing list thread, if one exist) and a solution is proposed.  This gives others the opportunity to discuss both the problem and the solution design before it's actually implemented.

laurencelundblade:
@thomas-fossati 

Before this change
- Only one nested EAT was possible (without duplicate map keys/labels; duplicating keys/labels is not good form).
- There was no way to name a nested EAT
- A nested EAT was treated separately from a submod so it made writing about inheritance and attachment in common harder.

draft-ietf-rats-eat.md
laurencelundblade: Oversight. Should be less than or equal too. Thanks for catching.

draft-ietf-rats-eat.md
laurencelundblade: Line 754 is explicit that there is no inheritance.

Agree that the wording here is fuzzy.

The nonce is intrinsically different from other claims as it does not describe the entity. It comes from the relying party to prevent replay. To replicate or not replicate the nonce could be considered to not be an inheritance question and we could have explicit text spelling all this out.

draft-ietf-rats-eat.md
laurencelundblade: Agreed

laurencelundblade:
Here's a comment from Ned that I'm copying to the mainline for easier reference:

> The submodule could have attestation capability itself in which case the verifier might be instructed to assess the subcomponent claims as another instance of an attestation token. Maybe it makes sense to include an attachment_type for attestable_submod? Additionally, the submodule could be attached using a secure session / protocol. I don't see this represented. The structure is an enumeration, but it is possible for multiple attachment_type conditions to be true at the same time. For example, a submodule could separately attests and be connected via secure session. Possibly attestable_submod is a property that belongs somewhere else? A secure_protocol attachment type could be true as well when one of the other attachment_types are true. Maybe it belongs somewhere else too? (Or these should be booleans).

Ned, it seems like you want to distinguish between "claims from a submod" and "attested claims from a submod". I don't know what the difference is (yet). Can you give some examples?

On attachment_type, we could make it an array of enumerated so you can specify more than one. We could even have an IANA registry for new types.  My thought is to try to come up with a small fixed useful set of classes. If we let 1000 flowers bloom here it will become useless to the relying party. They should be classes, rough categories, not exact types. The vendor has to pick the one most suited.

SimonFrost-Arm:
It would seem that 'attested claims from a submod' is already covered by nested-eat?

henkbirkholz:
Same problem as with PR #36 

If the CDDL fragment is supposed to be a type choice that represent a range of values, it is not doing that.

>     attachment_type = (
>     unspecified: 1,
>     device_internal: 2,
>     pcb_internal: 3,
>     chip_internal: 4
>     )
> 
>     attachment_type_claim = (
>     attachment: attachment_type )

If i understand the intent, I can add corresponding CDDL to the cddl-fixes branch.

SimonFrost-Arm:
I believe the appropriate CDDL would be
attachment_type_claim = &attachment_type
with attachment_type left as it is as a group. Per 2.2.2.2 of draft-ietf-cbor-cddl-03.

This encoding already appears to be in place in PR36

henkbirkholz:
Yes. https://tools.ietf.org/html/rfc8610#section-2.2.2.2 is an alternative use. Typically, it is used for either for a large number of values or if the group is also used in other contexts. In general, this is fine, too. It is just me being personally a bit more in favor of the style I just highlighted in #36.

PR #36 is broken nevertheless (sorry for the cross issue post), as the will not work, it  also uses string labels with implied cuts, and you cannot use a rule inside as an array item. I'll post another comment there.

Coming back to this PR. I assume what you want looks more like this (I'd also start with index "0" to be consistent with PR #36:

> attachment_type_claim = ( attachment => &attachment_type )
> 
> attachment_type = (
>   unspecified,
>   device_internal,
>   pcb_internal,
>   chip_internal,
> )
> 
> ;this is the CBOR specific value list
> unspecified = 0
> device_internal = 1
> pcb_internal = 2
> chip_internal = 3

laurencelundblade:
> It would seem that 'attested claims from a submod' is already covered by nested-eat?

This PR replaces the nested-eat claim.

nedmsmith:
Laurence said "Ned, it seems like you want to distinguish between "claims from a submod" and "attested claims from a submod". I don't know what the difference is (yet). Can you give some examples?"
Attested means signed by an Attester. If the submods/nested-eat allows for 'unsigned EAT' structures then it is possible to describe components that exist but are not 'attested'. There is a security semantic that could become misinterpreted by Verifiers if signing of something that is nested wasn't inspected by the Attester. Example is a peripheral device that is enumerated by BIOS but contains embedded firmware (as opposed to option-ROM). The BIOS is in the Attester but can't inspect the embedded firmware of the peripheral device. It can assert that the device exists (was enumerated) but can't vouch for the integrity of its firmware. The peripheral should be a second Attester based on a root of trust that exists in the peripheral. 

If the EAT structure describes both the peripheral device (D1) and the main board / platform (P1) then the Attester for P1 will likely sign an EAT that contains P1 claims (p0, p1, p2,...) and D1 claims (d0, d1, d2, d3,...) using the P1 attestation key. If [] denotes signing, the claim might look like [[p0, p1, p2, [d0, d1, d2,...]KD1 ]KP1; where K denotes a signing key. 

If on the other hand, there isn't an attester for D1, but P1 wants to assert that D1 exists. Then it would product an EAT with the structure [p0, p1, p2,..., d0, d1, d2, ...]KP1 

The point of possible confusion by the Verifier is whether or not P1 is saying that D1 exists or that D1 is attested by P1; where "attested" also (in addition to signing) means that the integrity of D1 is proven to be correct because of some special access P1 has to D1. 

The Interface (attachment_type) claim may be helpful in that it could describe the relationship between P1 and D1. For example, if D1 firmware were stored in memory that is dual ported to P1 then P1 could directly read it and compute its hash. However, the current interface claim structure doesn't describe the security relevant properties of the interface (AFAK).

I think the architecture intends for Attester to mean an environment that has special access to the Attestee such that it can prove its state is X AND that it can sign claims about X. 

If the structure of nested EATs somehow allows the Attester/Attestee relationships to breakdown then Verifiers will be confused about the security properties of what an EAT is asserting.

Issue #49, opened 2020-02-12
Disallow floating point time stamps in EAT?


laurencelundblade:
The nbf, exp and iat claims in CWT are allowed to be floating point. EAT inherits from them.

Floating point is not universally supported by CPUs, particularly small ones for IoT. Even if the CPU supports it, the compiler and OS might not (I’ve seen this happen once).

I don’t think fractional seconds are of any value for these claims, so we are forcing EAT decoders to support this for no good reason. I’m not sure why CWT allowed floating point for these claims other than just inheriting from the CBOR epoch time format.

It is possible to use SW to decode a float and turn it into an integral value, but that is not fun to write and seems a waste.

I’d like to specify in EAT that floating point is not allowed for these claims.

(Separately, we do need floating point for GPS location, but that claim will be used much less frequently and it is a good reason for using floating point).

laurencelundblade:
Resolved by #61

PR #61, opened 2020-04-20
Disallow floating-point dates

laurencelundblade:
CWT allows floating-point dates, but they are of no value and add an implementation burden, thus EAT is disallowing them.


laurencelundblade:
This closes #49

Issue #54, opened 2020-02-18
Accommodate "GP unendorsed" tokens


laurencelundblade:
Consider making an EAT a "CWT Claims Set" as defined in CWT (and JWT) to accommodate GP's desired for unsigned tokens.

laurencelundblade:
Fixed with #67 and the eventual publication of Henk's UCCS draft.

PR #62, opened 2020-04-21
Add defintion of UCCS and UJCS

laurencelundblade:
UCCS and UJCS are unprotected CWT/JWT for use in contexts where protection is unnecessary or provided by other means.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -56,6 +56,7 @@ author:
 
 normative:
   RFC2119:
+  RFC6838:
   RFC7049:
   RFC7519:
   RFC8126:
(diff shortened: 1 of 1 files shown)


jimsch:
The JOSE way to do this is (shudder) to create a JWT with the signature algorithm of None.  You probably need to review this depending on where you are planning to use this.

gmandyam:
Is this meant to address https://github.com/ietf-rats-wg/eat/issues/54?  If not, which issue is associated with the PR?

laurencelundblade:
Yes, this is to address #54

laurencelundblade:
Closed because UCCS is now in a separate draft by Henk

PR #67, opened 2020-10-26
submods: UCCS, bstr wrapping, tagging rules

laurencelundblade:
Allows UCCS tokens

Incorporates JWTs, secured and unsecured.

Requires nested CWT submodules be bstr wrapped.

How to bring a UCCS in as a submodule.

How to put a JWT submodule in a CWT and vice versa.

All the rules for CWT and COSE tag usage.

Top-level CDDL to handle CWT/COSE, UCCS and JWT


henkbirkholz:
I'd like to propose the following contributions that would introduce changes to the current proposal.

1.) The current root type is named eat-claims. I'd propose to change that to simply eat. eat-claims implies simply a claims set, I think.
2.) If you really want to include JWT, the nested-token type should look like this:

    nested-token = bstr .cbor eat / b64url

3.) Change the name "Non-token Submodules" to simply "Claims Sets" (label: nested-claims-set)
4.) Change the name "Nested EAT" to "Token" (label: nested-token). The nesting is intrinsic anyways and the term submodule does not really add anything - or maybe I am missing some context here? This also allows to "nest" CWT and not only EAT. CWT and EAT can only be differentiated by the composition of their Claims Sets anyways and therefore it is technically possible to use a CWT here already (although not allowed and that has to be checked by the application every time).
5.) In general, I wonder why submod-name is necessary. This way each "unknown" label has to assumed to be a nested token or claims set. That is rather complex to parse on the application layer, isn't it? I propose to use distinct and well-defined labels for CWT and CWT Claims Sets (and if we are really going to do this the JSON equivalents, still not convinced here).

laurencelundblade:
1) The most correct CDDL top level definition of an "eat" would be a type that is either a CWT, UCCS or JWT. It would reference UCCS and COSE cddl. It would take express the tag rules. However, that runs into quite a mess as there's no way I know of to say a "COSE with eat-claims as payload" and would require a COSE implementation inside the cddl tool to validate.

2) I tried that and got into trouble with the cddl validation tool. Also, are you sure JWT is always a complete and proper B64url? I think it has some semantically significant blank lines and there for is not.

3) I am open to name change, but want to get to common understanding first.

4) When something is an "EAT" its purpose is clearly attestation. When something is a "CWT" its purpose may be attestation or something else. Submods should only be tokens whose purpose is attestation.

5) The intended use of submod name is for things like "Acme TEE", "Qualcomm Hexagon DSP", "Infineon Secure Element". For truly standardized identification of a submod, OEMID, HWVersion and CoSWID should be relied upon by the RP.

draft-ietf-rats-eat.md
```diff
@@ -1432,6 +1492,19 @@ no new claims have been added.
 
 * Change IMEI-based UEIDs to be encoded as a 14-byte string
 
+<<<<<<< HEAD
```
thomas-fossati, line 1495: spurious line

PR #69, opened 2020-11-02
Factor out CDDL

thomas-fossati:
Reviewing the editor draft, I noticed the CDDL had a few errors and that the examples were drifting away. Therefore I have taken the liberty to do some restructuring to increase internal coherency, remove any need to copy&paste, and more generally automate all things around CDDL and examples.  Specifically:

* Extract CDDL for each claim into its own snippet file which is sourced from the markdown;
* Build the amalgamated CDDL from all the snippets automatically and source it from the markdown;
* Make the examples validate against the CDDL as a build/deploy precondition.

In the process I had to do the following pretty arbitrary moves:
* Fix the nonce claim to allow both array and scalar;
* Commented out `generic-claim-type` which was used without being defined;
* Consequently dropped from the examples the custom claims;
* Give EAT claims a code-point.

There is now a dependency on a couple of tools: `cddl` and `diag2cbor`.  I have made the Travis based CI work with them.  Hopefully Travis is the CI in use in the main repo?  If not, we'll need to port the changes to the target CI environment.

PS: the JSON claims have not been included yet in the full CDDL.


thomas-fossati:
> There is now a dependency on a couple of tools: `cddl` and `diag2cbor`. I have made the Travis based CI work with them. Hopefully Travis is the CI in use in the main repo? If not, we'll need to port the changes to the target CI environment.

I have made an image at https://hub.docker.com/repository/docker/thomasfossati/i-d-template that could be used for the circle CI.

I've also opened a PR on the upstream i-d-template: https://github.com/martinthomson/i-d-template/pull/250

PR #71, opened 2020-11-10
First draft of key inclusion

laurencelundblade:
Seems to me there are lots of use cases for key inclusion and they have semantics that vary too much to fit into  one claim. I’ve chosen primarily to write guidelines rather than define claims.

One use case is a CSR. However, RFC 8747 doesn’t define a CSR, so referencing it isn’t sufficient.

Attestations from the Android key store include things like security level and user authentication requirements for use. Going into all that is too much for EAT, but might be good in another document.

FIDO goes one step beyond Android and binds a relying party/user/device tuple with the key.

It is a little tempting to define a security level for the storage and operation of keys since EAT has a security level definition.


gmandyam:
See https://github.com/ietf-rats-wg/eat/pull/60.  Would recommend tying this concept in specifically for CSR.

laurencelundblade:
My thought is that a CBOR format CSR is of use outside of EAT and thus should be defined by the COSE work group and referenced by RATS.  A CBOR format CSR seems like a good and useful thing to do, but don't think it belongs in EAT. Maybe we hurry up and create it so EAT can reference it?

gmandyam:
>A CBOR format CSR seems like a good and useful thing to do, but don't think it belongs in EAT. Maybe we hurry up and create it so EAT can reference it?

I agree that a CBOR-formatted CSR would be a good idea, and I don't think we need to have an EAT dependency on it.  It can be done separately, as you suggested in the COSE WG.

Context claim, as intended by https://github.com/ietf-rats-wg/eat/pull/60, is not meant to convert or extend an EAT to a CSR object.  It is meant to describe how the attestation is intended to be used - e.g. in the context of a CSR.  I think that is sufficient for key attestation.  If a CBOR CSR object is defined, then it can encapsulate or be appended with an EAT with the appropriate context set.

nedmsmith:
Could someone summarize the use cases for how CSRs are used/useful? I'd like to understand how CSRs of a Evidence / Endorsement / Reference Value claim set fits into the attestation architecture.

henkbirkholz:
I fear that the "keys" topic introduces quite some scope creep to the core EAT I-D. I'd recommend to out-source these types of claims and a definition of their purpose to another complementary I-D.

Additionally, I think that CSR are not EAT at all. Having said that, it is absolutely fine to create a CBOR/CWT encoding for CSR, but I strongly recommend to do this in a separate I-D (and in a different one than above -> maybe not even in RATS but somewhere else?).

laurencelundblade:
@nedmsmith The use case I know of for including a CSR is IoT device on boarding. When the device first contacts some IoT management system, it includes a CSR. The management system checks out the attestation and maybe checks some other things not related to the attestation and verifies the possession signature in the CSR. If all is OK, the management system records the key in the CSR and uses it with DTLS for further communication with the device.

This isn't my design. I just know some folks are interested in this.

So it's not really a CSR since no signed certificate is returned. The CSR is being used because it provides proof of possession that just including a bare key does not.

At any rate, it seems we're agreed that we will not do any work in EAT to either define a CBOR CSR or to define a claim that allows for their inclusion. (but of course no prohibition against them should some one feel the need)

laurencelundblade:
Now that we agree that CSRs are off the table in EAT, what do you think of the proposed text?

I noticed that Google/Android prototype codes puts keys in EAT, but not as a COSE_Key.  The text seems like a good hint to not do things like that.

draft-ietf-rats-eat.md
```diff
@@ -739,9 +739,9 @@ An EAT may include a cryptographic key such as a public key.
 The signing of the EAT binds the key to all the other claims in the token.
 
 The purpose for inclusion of the key may vary by use case.
-For example, the key may be included to request certificate signing (a CSR).
+For example, the key may be included as part of an IoT device on boarding protocol.
```
thomas-fossati, line 742:
```suggestion
For example, the key may be included as part of an IoT device onboarding protocol.
``

PR #72, opened 2020-11-14
Add CDDL for JSON nonce label; indicate CDDL from CWT, JWT... is included in EAT

laurencelundblade:


draft-ietf-rats-eat.md
```diff
@@ -851,6 +851,9 @@ string naming the submodule. No submodules may have the same name.
 # Encoding {#encoding}
 This makes use of the types defined in CDDL Appendix D, Standard Prelude.
 
+Some of the CDDL included here is for claims that defined in CWT {{RFC8392}} or JWT {{RFC7519}} or are in the IANA CWT or JWT registries.
```
thomas-fossati, line 854:
```suggestion
Some of the CDDL included here is for claims that are defined in CWT {{RFC8392}} or JWT {{RFC7519}} or are in the IANA CWT or JWT registries.
```

cddl/json.cddl
```diff
@@ -1,4 +1,5 @@
 ueid /= "ueid"
+nonce /= "nonce"
```
thomas-fossati, line 2: Laurence, while you are on it:
* `debug-disble` should be `debug-disable`, and
* `nested-eat` should be deleted since it doesn't seem to be used

thomas-fossati: Also, I think now we could uncomment `CDDL_FRAGS += json.cddl` in `cddl/vars.mk`

PR #73, opened 2020-11-14
Clarify that the security level claim is NOT extensible

laurencelundblade:
If everyone made up their own security level as they will be wont to do, this claim would be generally useless.

See also #4


draft-ietf-rats-eat.md
```diff
@@ -621,9 +645,13 @@ against physical or electrical attacks against the device itself.
 It is assumed any potential attacker has captured the device and can 
 disassemble it. Example include TPMs and Secure Elements.
 
+The entity should claim the highest security level it achieves and no higher.
+This set is NOT extensible so as to provide a clear common interoperable description of security level to the relying party.
+If a particular implementation considers this claim to be inadequate, they should 1) omit the claim out, 2) define their own proprietary claim or 3) both.
```
thomas-fossati, line 650: I would simply say:
_"If a particular implementation considers this claim to be inadequate, they should omit it and define their own proprietary claim instead."_

laurencelundblade:
Part address #74

PR #81, opened 2020-11-16
First proposal for measurements claim

laurencelundblade:


SimonFrost-Arm:
The use case for this claim seems to be relevant where a large(?) set of measurements are required for a single purpose e.g. Linux IMA as cited.  Where a single measurement accompanies separate metadata for the context of use (e.g. as part of a CoSWID), this structure is not helpful in its current form (too complex for that purpose). It may be better to separate the use cases rather than have a single measurement type. Perhaps therefore this claim should be called 'bulk_measurement' or similar.

laurencelundblade:
@SimonFrost-Arm

I think this could be used for simple measurements too. Here's an example of a single measurement.

    25:                       // label/key indicating measurements claim
      [                       // Array of one measurement scheme
        {                     // Map containing a measurement
          0: "Acme integrity" // The measurement scheme
          2: "HLOS Kernel"    // What was measured
          3:                  // The measurement values
            [                 // Array of one measurement value
              "code text": h'459a7e9820a8749a8a123410c8' // The one measurement value
            ]
        }
      ]

I think this parallels CoSWID in some ways. Some of the structures are a little complex, but mostly reduce in the simple cases.

Can you describe the simple measurements of interest?

Also, this probably needs an optional timestamp and age field as the measurements may be taken at a time different from the token generation.

SimonFrost-Arm:
Measurement is probably not standalone in usage, but a sub-part of a wider object describing software. If I'm already describing an object, but want to include a measurement, I have probably already identified it and don't need to repeat the measurement-target entry. 

Measurement-name (NB renaming this to measurement-scheme may avoid confusion with measurement-target, I had to look twice to confirm which was which) and Measurement-value-name may be implicit for the type of use. eg. in a deployment all measurements may use SHA256 and always measure the code text. If I state this in the profile for the deployment, I shouldn't need explicit text strings to accompany the measurement. 

I'll put the usual appeal in for constrained environments: even with single character text entries, this adds 17 bytes overhead to every measurement. With the example quoted, there is a 45 byte overhead. 

Can we make all of the explanatory fields accompanying the actual measurement optional? That would allow more flexible use and probably address the above concerns.

laurencelundblade:
Helpful comments. Thanks Simon. Will give it another try (but will probably focus on some other work to get a new draft out in the short term).

laurencelundblade:
This is probably replaced by #102

laurencelundblade:
Addressed by #102

PR #86, opened 2020-12-09
Add profile claim and section

laurencelundblade:


SimonFrost-Arm:
I approve of the addition of this claim. The description covers what should be in the resultant doc well and is consistent with the example use of this claim in the Arm PSA profile document.

PR #93, opened 2021-02-25
Change profile claim to be URL or OID

laurencelundblade:


cddl/profile.cddl
```diff
@@ -1,3 +1,3 @@
 profile-claim = (
-    profile => tstr
+    profile => tstr / oid
```
thomas-fossati, line 2: CDDL nits: I think we need to import the definition of oid from https://datatracker.ietf.org/doc/draft-ietf-cbor-tags-oid/, i.e.: `oid = #6.111(bstr)`

Also, if we are requiring an URL, maybe better letting the type system help out: `uri = #6.32(tstr)`?

draft-ietf-rats-eat.md
```diff
@@ -954,9 +954,12 @@ security state of the entity storing the private key used in a PoP application.
 
 ## The Profile Claim (profile) {#profile-claim}
 
-The profile claim is a text string that simply gives the name of the profile to which the token purports to adhere to.
-It may name an IETF document, some other document or no particular document.
+A profile is identified by either a URL or an OID.
+Typically, the URL will reference a document describing the profile.
+An OID is just a unique identifier for the profile.
+It may exist in anywhere in the OID tree.
```
thomas-fossati, line 960:
```suggestion
It may exist anywhere in the OID tree.
```

draft-ietf-rats-eat.md
```diff
@@ -954,9 +954,12 @@ security state of the entity storing the private key used in a PoP application.
...
 There is no requirement that the named document be publicly accessible.
```
thomas-fossati, line 961: I think this should be moved up to L959

thomas-fossati (commented):
Thanks for taking a stab at this!

cddl/profile.cddl
```diff
@@ -1,3 +1,3 @@
 profile-claim = (
-    profile => tstr / oid
+    profile => ~uri / ~oid
```
thomas-fossati, line 2: a curiosity: what is the rationale for unwrapping?

laurencelundblade: You can tell uri from old by the CBOR type (text vs byte string) so no need for the tag. 

Generally, it seems the convention in CBOR is to unwrap if possible to not have redundancy (which creates an error condition) and to save bytes. If others think otherwise, fill me in. :-)

thomas-fossati: makes sense, thanks for the explanation

draft-ietf-rats-eat.md
```diff
@@ -954,11 +963,20 @@ security state of the entity storing the private key used in a PoP application.
...
 ## The Profile Claim (profile) {#profile-claim}
 
-The profile claim is a text string that simply gives the name of the profile to which the token purports to adhere to.
-It may name an IETF document, some other document or no particular document.
+See {{profiles}} for the detailed description of a profile.
+
+A profile is identified by either a URL or an OID.
+Typically, the URL will reference a document describing the profile.
+An OID is just a unique identifier for the profile.
+It may exist anywhere in the OID tree.
```
thomas-fossati, line 971:
```suggestion
```
these two lines are duplicates of L973-974

draft-ietf-rats-eat.md
```diff
@@ -954,11 +963,20 @@ security state of the entity storing the private key used in a PoP application.
 
 ## The Profile Claim (profile) {#profile-claim}
 
-The profile claim is a text string that simply gives the name of the profile to which the token purports to adhere to.
-It may name an IETF document, some other document or no particular document.
+See {{profiles}} for the detailed description of a profile.
+
+A profile is identified by either a URL or an OID.
```
thomas-fossati, line 968:
```suggestion
A profile is identified by either an absolute URI {{!RFC3896}} or an absolute OID (ASN.1 Type "OBJECT IDENTIFIER").
```

draft-ietf-rats-eat.md
```diff
@@ -954,11 +963,20 @@ security state of the entity storing the private key used in a PoP application.
...
+Typically, the URL will reference a document describing the profile.
```
thomas-fossati, line 969:
```suggestion
Typically, the URI will reference a document describing the profile.
```

draft-ietf-rats-eat.md
```diff
@@ -954,11 +963,20 @@ security state of the entity storing the private key used in a PoP application.
...
+An OID is just a unique identifier for the profile.
+It may exist anywhere in the OID tree.
 There is no requirement that the named document be publicly accessible.
+An OID is just a unique identifier for the profile.
+It may exist anywhere in the OID tree.
```
thomas-fossati, line 974: I don't understand why is it important to state this?

draft-ietf-rats-eat.md
```diff
@@ -954,11 +963,20 @@ security state of the entity storing the private key used in a PoP application.
...
+The primary purpose of the profile claim is to uniquely identify the profile even if it is a private profile.
 
-See {{profiles}} for a detailed description of a profile.
+The OID is encoded in CBOR according to {{CBOR-OID}} and the URI according to {{RFC8949}}.
+Both are unwrapped and thus not tags.
```
thomas-fossati, line 978:
```suggestion
Both are unwrapped and thus the outer tag is removed.
```

draft-ietf-rats-eat.md
```diff
@@ -1244,6 +1262,8 @@ following CDDL types are encoded in JSON as follows:
 * bstr -- must be base64url encoded
 * time -- must be encoded as NumericDate as described section 2 of {{RFC7519}}.
 * string-or-uri -- must be encoded as StringOrURI as described section 2 of {{RFC7519}}.
+* uri -- must be a URI {{RFC3986}}.
+* oid -- is encoded as a string represetation of the OID digits.
```
thomas-fossati, line 1266:
```suggestion
* oid -- is encoded using dotted decimal notation as per Section 5 of {{CBOR-OID}} 
```

draft-ietf-rats-eat.md
```diff
@@ -132,6 +133,14 @@ normative:
     - fullname: B. de Medeiros
     - fullname: C. Mortimore
 
+  CBOR-OID:
+    target: https://tools.ietf.org/html/draft-ietf-cbor-tags-oid-05
+    title: Concise Binary Object Representation (CBOR) Tags for Object Identifiers
+    date: 2021
+    author: 
+    - fullname: C. Bormann
```
thomas-fossati, line 141:
```suggestion
  CBOR-OID: I-D.ietf-cbor-tags-oid
```

thomas-fossati: so that updates to the ref are tracked automatically

draft-ietf-rats-eat.md
laurencelundblade: This is JSON representation of OIDs so a reference to CBOR representation of OID seems wrong.
I haven't found any standard to reference for text representation of OIDs in JSON so far.

thomas-fossati: Yeah that doesn't sound entirely right, but it's the only reference I had found.

I've given my google-foo another spin and found that Section 3.2 of RFC5675 has an ABNF for the dotted-decimal notation.  Or otherwise let's just say "using the well established dotted-decimal notation" or similar

draft-ietf-rats-eat.md
laurencelundblade: I'm leaving this as is because, there is no possibility of any tag here (no inner tag). Also, contrary to normal English use of the word tag, in CBOR the tag is the whole thing with the value, not just the tag number.

draft-ietf-rats-eat.md
laurencelundblade: OK. Pushed a change for this.

PR #105, opened 2021-04-10
First attempt at claims for layered attestation

laurencelundblade:
Relates to https://github.com/ietf-rats-wg/architecture/issues/293 and https://github.com/ietf-rats-wg/architecture/pull/294

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -919,7 +919,7 @@ It is not the same as a seed for a random number generator which must be kept se
 {::include cddl/boot-seed.cddl}
 ~~~~
 
-## The Intended Use Claim (intended-use)
+## The Intended Use Claim (intended-use) {#intended-use}
 
 EAT's may be used in the context of several different applications.  The intended-use
 claim provides an indication to an EAT consumer about  the intended usage
@@ -951,6 +951,10 @@ proof-of-possession (PoP) appication. More precisely, a PoP transaction is inten
 to provide to the recipient cryptographically-verifiable proof that the sender has posession
 of a key.  This kind of attestation may be neceesary to verify the
 security state of the entity storing the private key used in a PoP application.
+
+6 -- Attesting Environment
+: This is used only in submodules. It indicates that the target of the 
+Attestation Evidence is itself an Attesting Environement. See {{layered}}.
 
 ### intended-use CDDL
 
@@ -1099,11 +1103,52 @@ string naming the submodule. No submodules may have the same name.
 {::include cddl/submods.cddl}
 ~~~~
 
+
 # Endorsements and Verification Keys {#keyid}
 
 TODO: fill this section in. It will discuss key IDs, endorsement ID and such that
 are needed as input needed to by the Verifier to verify the signature. This will
 NOT discuss the contents of an Endorsement, just and ID/locator.
+
+
+# Layered Attestation {#layered}
+
+It is possible for a device to have multiple Attesting Environments as described in {{RATS-Architecture}}.
+When these are layered, one is a root the endorsement and/or verification key originates with the device manufacturer or endorser.
+The trust for those that are layered above the root comes from Attestation Evidence about them that is provided by the root or a layer below.
+
+A layered Attesting Environment is always described as a submodule.
+To know that a particular submodule is providing evidence for an Attesting Environment, it must have an intended use claim with the value of "attest=env".
+See {{intended-use}}.
+
+The submodule should also contain a claim that hold the verification key, the key used to verify signed attestation evidence from the layered attester.
+A specific claim is defined to hold verification keys.
+There is no requirement on the algorithm or key type used by the layered attestation.
+There is a requirement that the included key is in the format of a COSE_key.
+
+There are two ways that the Verifier comes to trust a layered Attesting Environment.
+
+First, the root or lower layer Attesting Environment provides claims, perhaps many claims, characterizing the Attesting Environment.
+The Verifier recognized the claims are about an Attesting Environment by the intended use claim and uses the other claims to decide whether it trusts this layered Attesting Environment.
+The decision is made by the Verifier.
+
+Second, the root or lower layer Attesting environment provides the is-trusted claim with a value of true.
+This tells the Verifier that the Attesting Environment has been sufficiently evaluated on the device by the root or lower layer Attesting Environment.
+The decision is made by on the device.
+
+Once the Verifier knows the Attesting Environment is trusted, then it makes use of the verification key to verify Attestation Evidence from the layered Attesting Environment.
+
+Multiple layers of Attesting Environments are allowed and must be represented as nested submodules.
+The layered Attesting Environment must be a submodule of the root of the token or a lower layer Attesting Environment.
+It doesn't have to be an immediate submodule, though. 
+It can be a submodule of a submodule as long as none of the intermediate submodules are not layered Attesting Environments.
+
+
+~~~~CDDL
+{::include cddl/layered.cddl}
+~~~~
+
+
 
 # Profiles {#profiles}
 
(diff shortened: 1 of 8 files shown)


draft-ietf-rats-eat.md
```diff
@@ -952,6 +952,10 @@ to provide to the recipient cryptographically-verifiable proof that the sender h
 of a key.  This kind of attestation may be neceesary to verify the
 security state of the entity storing the private key used in a PoP application.
 
+6 -- Attesting Environment
+: This is used only in submodules. It indicates that the target of the 
+Attestation Evidence is itself an Attesting Environement. See {{layered}}.
```
henkbirkholz, line 957: Not the current Attesting Environment that generated that Evidence, tough, but a "potential future" one in an upcoming "secure state transition" to the next layer. Maybe this could be read as "an Attesting Environment generates Evidence about itself", which it does not do?

draft-ietf-rats-eat.md
```diff
@@ -1099,12 +1103,53 @@ string naming the submodule. No submodules may have the same name.
...
 TODO: fill this section in. It will discuss key IDs, endorsement ID and such that
 are needed as input needed to by the Verifier to verify the signature. This will
 NOT discuss the contents of an Endorsement, just and ID/locator.
 
+
+# Layered Attestation {#layered}
+
+It is possible for a device to have multiple Attesting Environments as described in {{RATS-Architecture}}.
+When these are layered, one is a root the endorsement and/or verification key originates with the device manufacturer or endorser.
+The trust for those that are layered above the root comes from Attestation Evidence about them that is provided by the root or a layer below.
```
henkbirkholz, line 1118: This addresses the concern above, but later in the text. Maybe some minor rewording above suffices.

draft-ietf-rats-eat.md
```diff
@@ -1099,12 +1103,53 @@ string naming the submodule. No submodules may have the same name.
...
+
+A layered Attesting Environment is always described as a submodule.
```
henkbirkholz, line 1120: so... initially there is the root Attesting Environment and state transitions then go to layered Attesting Environments? I kinda like these words. Will bring that up in the arch team.

draft-ietf-rats-eat.md
```diff
@@ -1099,12 +1103,53 @@ string naming the submodule. No submodules may have the same name.
...
+To know that a particular submodule is providing evidence for an Attesting Environment, it must have an intended use claim with the value of "attest=env".
+See {{intended-use}}.
+
+The submodule should also contain a claim that hold the verification key, the key used to verify signed attestation evidence from the layered attester.
```
henkbirkholz, line 1124: That's impossible for me to parse. The current Attesting Environment should include the pub-key of a layered Attesting Environment in the Evidence generated about that layered Attesting Environment? If so, maybe elaborate on the reasoning? What's the use case you have in mind here?

draft-ietf-rats-eat.md
```diff
@@ -1099,12 +1103,53 @@ string naming the submodule. No submodules may have the same name.
...
+
+There are two ways that the Verifier comes to trust a layered Attesting Environment.
+
+First, the root or lower layer Attesting Environment provides claims, perhaps many claims, characterizing the Attesting Environment.
+The Verifier recognized the claims are about an Attesting Environment by the intended use claim and uses the other claims to decide whether it trusts this layered Attesting Environment.
+The decision is made by the Verifier.
+
+Second, the root or lower layer Attesting environment provides the is-trusted claim with a value of true.
+This tells the Verifier that the Attesting Environment has been sufficiently evaluated on the device by the root or lower layer Attesting Environment.
+The decision is made by on the device.
```
henkbirkholz, line 1137: So these are "EAT options", right? In general, a third option is that the Verifiers can obtain Endorsements that are about the Attesting Environment, I think.

draft-ietf-rats-eat.md
```diff
@@ -1099,12 +1103,53 @@ string naming the submodule. No submodules may have the same name.
...
+
+Once the Verifier knows the Attesting Environment is trusted, then it makes use of the verification key to verify Attestation Evidence from the layered Attesting Environment.
+
+Multiple layers of Attesting Environments are allowed and must be represented as nested submodules.
+The layered Attesting Environment must be a submodule of the root of the token or a lower layer Attesting Environment.
+It doesn't have to be an immediate submodule, though.
```
henkbirkholz, line 1143: I don't understand what that means. It probably means "The layered Attesting Environment" from the first sentence, but still I cannot really make sense of it.

draft-ietf-rats-eat.md
```diff
@@ -1099,12 +1103,53 @@ string naming the submodule. No submodules may have the same name.
...
+Second, the root or lower layer Attesting environment provides the is-trusted claim with a value of true.
+This tells the Verifier that the Attesting Environment has been sufficiently evaluated on the device by the root or lower layer Attesting Environment.
+The decision is made by on the device.
+
+Once the Verifier knows the Attesting Environment is trusted, then it makes use of the verification key to verify Attestation Evidence from the layered Attesting Environment.
+
+Multiple layers of Attesting Environments are allowed and must be represented as nested submodules.
+The layered Attesting Environment must be a submodule of the root of the token or a lower layer Attesting Environment.
+It doesn't have to be an immediate submodule, though. 
+It can be a submodule of a submodule as long as none of the intermediate submodules are not layered Attesting Environments.
```
henkbirkholz, line 1144: Same

draft-ietf-rats-eat.md
nedmsmith: Regarding lines 955 - 957: The target environment contains an attesting environment but could contain other code, setup vectors, files or code that has functionality other than that used by an attesting environment. Structures representing a layer 'target environment' should allow for multiple disparate measurements.

draft-ietf-rats-eat.md
nedmsmith: 1117 seems to be saying "An Endorser provides claims about the root since there isn't a layer beneath the root layer that supplies Evidence. An Endorser also provides cryptographic material used by the root to authenticate the device and to sign evidence pertaining to upper / other layers."

draft-ietf-rats-eat.md
nedmsmith: 1121 seems too strong. The names of the measurements pertaining to the attesting environment (or other measurements in the target) provides intended use semantics. If using submods then submod-name captures the intended semantics.

1124 seems to be saying layers can be cryptographically linked, therefore it may be appropriate to include a public key in the Evidence related to an attesting environment.

draft-ietf-rats-eat.md
nedmsmith: 1135 I don't see how it is possible for a lower layer to know or assert that it knows the upper layer is trusted without going through a recursive verification. However, verification is done by a verifier which is a complex function. It would be a mistake to assert that layering requires attesting environments to also be verifiers according to the RATS Arch. The is-trusted claim is actually an attestation result structure. Doesn't belong here IMO.

draft-ietf-rats-eat.md
nedmsmith: There are two ways to structure layers. (a) as nested entities or (b) as linked entities. Nesting has the undesirable side-effect that the signer (aka top layer) has to be in possession of all of the lower layer evidence in order to sign anything. This makes for ever increasing evidence blobs. Linking allows the upper layer to only sign the evidence it directly controls. The lower layer linked its evidence about the upper layer so there is no reason to counter-sign something that has already been signed by a lower layer.  Lets not think of submods as a hammer that forces nesting in ways that don't make sense.

draft-ietf-rats-eat.md
nedmsmith: 1141 this is too strong.

laurencelundblade:
No consensus on this. If this gets done it should probably be in a future draft.

PR #124, opened 2021-08-19
Make OEM ID specifically for HW, add PEN option, add RNG option

laurencelundblade:
Previously OEM ID was just an IEEE OUI.

Now it is specifically for identifying HW.

In addition to IEEE, PEN and a 128-bit random number of a hash of something can be used.

### DIFF
--- a/cddl/oemid.cddl
+++ b/cddl/oemid.cddl
@@ -1,3 +1,10 @@
+oemid-pen = int
+
+oemid-ieee = bstr .size 3
+
+oemid-random = bstr .size 128
+
+
 oemid-claim = (
-    oemid => bstr
+    oemid => oemid-random / oemid-ieee / oemid-pen
 )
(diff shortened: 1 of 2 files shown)


nedmsmith (commented):
If you make oemid a type choice (ie $oemid) the parser can more deterministically resolve which form to use.  e.g.:

```
$oemid /= oemid-pen
$oemid /= oemid-ieee
$oemid /= oemid-random
```

If there is a strong desire to allow URIs as vendor IDs that could be supported in the future as
`$oemid /= uri `
or
`$oemid /= tstr`

laurencelundblade:
Would not want URIs. If you had some set of URIs you wanted to use, they should be hashed and go in oemid-random.

I also don't really want to encourage any more forms for oemid for the sake of simplicity and to pin vendors into doing one of the approved schemes which we know is thought out and works correctly to give global uniqueness.

This is definitely not about allowing every vendor, standards org and specific industry to plug in their existing company naming schemes. It's about picking a minimal set that works.

Issue #138, opened 2021-11-08
Meet TEEP requirements for non-TEEP/SUIT specific claims


dthaler:
https://datatracker.ietf.org/doc/html/draft-ietf-teep-architecture#section-7 and 
https://datatracker.ietf.org/doc/html/draft-ietf-teep-protocol#section-4.3.1 have requirements to be able to somehow express, in some fashion (ideally not TEEP specific):

- [X] Device unique identifier - I think "ueid" meets this, if I understand correctly (and this was mentioned in the mailing list thread)
- [X] Vendor of the device - I think "oemid" can now do this
- [ ] Class of the device
- [X] TEE firmware type - I think "software-name" can now do this
- [X] TEE firmware version - I think "software-version" can now do this

https://datatracker.ietf.org/doc/draft-birkholz-rats-suit-claims/ proposed claims that are not TEEP/SUIT specific for all of the above:
* vendor-identifier, class-identifier, device-identifier, component-identifier, version

The TEEP protocol spec has a normative dependency on those, and we agreed they are not TEEP or SUIT specific.   This issue requests adding them, or some other way of expressing similar information, into the base EAT spec.  This has been discussed on the RATS mailing list and asked for in multiple past RATS meetings, so now tracking via this issue.

The latest version of draft-ietf-rats-eat added software-name and software-version which I believe are sufficient to meet the firmware type and version requirements, so this now just tracks anything remaining, which I think is just device class.

dthaler:
Here's some of the discussion links, there are more (earlier ones) I can dig up later if needed:

* RATS email thread: https://mailarchive.ietf.org/arch/msg/rats/NRVt_SCZ7yH8M3SRbUZqlO448MY/
* Slides from IETF 111: https://datatracker.ietf.org/meeting/111/materials/slides-111-rats-sessc-rats-session-2-3-slide-bundle-03 slides 53-59
* Notes from IETF 111: https://notes.ietf.org/notes-ietf-111-rats

dthaler:
The relevant part of draft-birkholz-rats-suit-claims that is not specific to SUIT or TEEP is:

> 3.1.  System Properties Claims
> 
>    System Properties Claims are composed of:
> 
>    *  Hardware Component Claims and
> 
>    *  Software Component Claims.
> 
>    Correspondingly, the Claim definitions below highlight if a Claim is
>    generic or hw/sw-component specific.
> 
> 3.1.1.  vendor-identifier
> 
[...]

PR #139, opened 2021-11-08
Add HW Class Claim

laurencelundblade:


laurencelundblade:
Address #138

cddl/hardware-class.cddl
```diff
@@ -0,0 +1 @@
+$$claims-set-claims //= (hardware-class-label => bytes)
```
thomas-fossati, line 1: is the assumption here that the raw `bytes` is really `bytes .cbor pluggable_hw_class_type` ?

laurencelundblade: My thinking is that the structure of the bytes is OEM-specific and that we should not try to herd the OEM's into any standard formats. Doesn't seem like any need for this.

An OEM could define the structure in CBOR and then use .cbor to write CDDL for it. Nothing we're doing here stops that.

Kind of the intent here is the concatenation of HW OEM ID, HW Class and HW Version really gets you the exact unique type of the HW. You can do that without knowing the structure of the individual claims.

So, yes, really a byte string. (I thought about also allowing a text string)

thomas-fossati: Noting that this approach makes the claim extremely context sensitive.

I guess those EATs that carry an explicit profile claim, that could be used to unambiguously cast to the intended format.

gmandyam:
This is blocking the next release of the spec.  Can we merge?

cddl/hardware-class.cddl
henkbirkholz: I am having a hard time understanding how to use this claim. Maybe we can have a comprehensive example from the TEEP space that demonstrates very precisely how this claim (and the other "teep required claims" are supposed to work?

henkbirkholz:
I do not see consensus on this, I am afraid. Could we please do an explicit call for consensus on the list real quick? Like "this is how it works in English test, here are three examples in JSON and CBOR, any objections?"

cddl/hardware-class.cddl
laurencelundblade: @thomas-fossati  I think this works without any specified structure if the receiver understands it is only to provide uniqueness.  The original specification of this from Henk has no structure either https://datatracker.ietf.org/doc/html/draft-birkholz-rats-suit-claims-02#section-3.1.2

I'm not super opposed to structure if there's a clear reason for it, but it seems overkill right now.

laurencelundblade: @henkbirkholz This is a replacement for https://datatracker.ietf.org/doc/html/draft-birkholz-rats-suit-claims-02#section-3.1.2 so it should be used in a similar way.

I probably could add some text saying it can be thought of as a HW model number.

thomas-fossati: > @thomas-fossati I think this works without any specified structure if the receiver understands it is only to provide uniqueness.

Reading the accompanying prose below, I gather the assumption is that this is a completely unregulated space, so I am not sure how uniqueness is guaranteed?

> The original specification of this from Henk has no structure either https://datatracker.ietf.org/doc/html/draft-birkholz-rats-suit-claims-02#section-3.1.2 It is a UUID which has some structure but don't think the UUID structure is the same as the structure you are after.

Not a lot of extensibility there, but at least it's a standardised way to mint *unique* IDs. 

> I'm not opposed to structure if there's a clear reason for it, but it seems overkill right now and I'm not sure what that structure would look like.

I agree.  One way to attack this would be what I wrote [upthread](https://github.com/ietf-rats-wg/eat/pull/139/files#r745651422).

laurencelundblade: My interpolation of the TEEP/SUIT requirements  is that it up to the OEM to create what ever uniqueness it thinks is necessary under their HW OEM ID. Some might want a lot, some might not want any.

From the long threads with Brendon and TEEP/SUIT in the past, it seems that this is really about the OEM choosing what uniqueness they want for their product lines, their SW loads and such.

While UUIDs give some uniqueness techniques, they don't inform the SUIT/TEEP use case because UUIDs are not about product lines and SW load compatibility.

Do you have any examples of what formats you'd define for use with a socket/plug? That would really help me understand.

I'm trying to interpolate the use case and come up with a design to move EAT forward. I am also OK deleting this claim or making it TEEP/SUIT specific. I am kind of a middle-man here.

thomas-fossati: > My interpolation of the TEEP/SUIT requirements is that it up to the OEM to create what ever uniqueness it thinks is necessary under their HW OEM ID. Some might want a lot, some might not want any.

Lifting this up into EAT automatically gives it a much wider applicability, so we need to carefully look into the implications.  In its current form I don't think this can work unless we [couple it to the profile claim](https://github.com/ietf-rats-wg/eat/pull/139/files#r745959664).  But if that is the case, I don't see a strong reason for giving it global semantics.

> From the long threads with Brendon and TEEP/SUIT in the past, it seems that this is really about the OEM choosing what uniqueness they want for their product lines, their SW loads and such.

Thanks a lot for the additional context.

> While UUIDs give some uniqueness techniques, they don't inform the SUIT/TEEP use case because UUIDs are not about product lines and SW load compatibility.
> 
> Do you have any examples of what formats you'd define for use with a socket/plug? That would really help me understand.

I'd start with a set of standardised identifiers that either have strong collision-resistant properties, or are natively namespaced.   So, the first three that come to mind are OIDs, URIs, UUIDs.  And, most importantly, I'd make it possible to extend the type.  Concretely:
```
hardware-class-label => $hardware-class-type

$hardware-class-type /= tagged-oid
$hardware-class-type /= tagged-uuid
$hardware-class-type /= tagged-url

; extensions will need to allocate their own CBOR tag, e.g., in another document:
$hardware-class-type /= tagged-vendor-x-hw-class-type
```
 
> I'm trying to interpolate the use case and come up with a design to move EAT forward. I am also OK deleting this claim or making it TEEP/SUIT specific. I am kind of a middle-man here.

Appreciate this fully; apologies for taking on the role of the torturer. :-)

bremoran:
Hardware classes MUST NOT be human readable. If they are, they will be abused as human-facing identifiers, then incompatible hardware revisions will be shoehorned into the same "version" for the sake of "preserving image," hiding cost reductions, etc., etc.

Hardware classes are for machines to read. Don't make them human readable. Humans don't need to read them. UUIDs are a great solution because you can't read them, so you won't be tempted to hide human-readable information in them. A plain hash is good too.

cddl/hardware-class.cddl
quic-jodonogh: It is both reasonable for EAT to define a base set of types and to allow these to be extended by profiles, accepting that profile-defined types inevitably introduce some context sensitivity.

However I don't think we need to boil the ocean at this time - it would be a massive undertaking to determine a universal (or even "fairly widely applicable") HW class type, and I don't think we ned to do this in EAT.

If I look at the GlobalPlatform TEE identifiers, for example (a reasonable class of TEE to wish to attest even if not at all TEEP compliant), I see no reasonable way to do something that would be interoperable outside of a GlobalPlatform profile (where interoperability would be assured).

Looking at Thomas' approach above, only problem I have is what a verifier should do if it sees a `$hardware-class-type` it does not understand. For this reason I think I prefer something like:

```
hardware-class-label => $hardware-class-type

$hardware-class-type /= tagged-oid
$hardware-class-type /= tagged-uuid
$hardware-class-type /= tagged-url
$hardware-class-type /= bytes .cbor pluggable-x-hw-class-type
```

which has some similarity to Thomas' first comment. It would be straightforward for e.g. GlobalPlatform to define its own HW class type.

I'm not sure that it is absolutely necessary to tie use of pluggable HW class type to the profile claim with a SHALL, but it seems reasonable to me that pluggable class type SHOULD be used in conjunction with profile claim and that the combination profile claim + CBOR encoded pluggable class type MUST uniquely distinguish different HW.

quic-jodonogh (approved):
This approach makes sense to me. EAT Profiles can always make the format of the byte strong explicit if they wish to do so.

PR #141, opened 2021-11-25
Reduce HW Version to one claim

laurencelundblade:
Rather than having separate claims for HW Version for chip, board and device, there is only one. 

An example is given using sumods to achieve what was possible before with the three claims, including giving OEM IDs and OEM classes.


quic-jodonogh (approved):
Verified - all looks as we discussed today, and sufficient to meths e requirement. Much preferred over the previous chip/board/device labels.

Example for board + device submods is helpful

PR #147, opened 2021-12-22
Add HW Class claim (really this time)

laurencelundblade:


laurencelundblade:
This is to really address #138

laurencelundblade:
Here's some design rationale...

From talking to Dave, this is really about HW IP like that from Arm, Synopis and the like. These are companies that do HW designs, but don't manufacture them. With contract foundries like TSMC and these HW IP designers, there's lots more people making chips with a lot more common HW blocks, so I can see the use in such a generic claim.

Since I worked on HW for a while, I can say "IP" is the term used in the HW industry. Any interest in calling this the "HW IP Claim"? It would be crystal clear to me, where as "Class" was extremely confusing to me.
https://en.wikipedia.org/wiki/Semiconductor_intellectual_property_core

The identifier really does have to be globally unique. It can't be anchored with the OEM, because the key point here is that multiple OEMs are using the same design. It can't be anchored in something like a MAC address or IEEE OUI.

The two extant global identifier schemes not associated with an OEM are URI's and OIDs, so I've allowed them.

Global uniqueness can also be achieved probabilistically with a large random number, so there's a 16-byte string too. There's probably not very many of these designs, only thousands, so 16 bytes is enough. This also fulfills the non-human-readable characteristic that is desired. It can be a UUID too.

For this to work, someone has create an ID for each HW design and make sure the OEMs manufacturing with the design use that ID. Ideally, each HW designer would tell the OEM the EAT HW Class for the HW block you are putting in your chip is xxx.

I thought about an IANA registry for HW IP identifiers, but that seems a bit of an overreach. On the other hand, it would be clear what they all are and they could be integers like PENs. They would be very compact.

laurencelundblade:
A HW model number was added. There was no consensus on HW Class/IP.

PR #148, opened 2021-12-27
Major change to prefer the term "entity" to "device"

laurencelundblade:
Address #16 and other comments.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -243,9 +243,9 @@ The claims defined in this document are claims about an entity.
 
 An entity is an implementation in hardware, software or both.
 
-An entity is the same as an Attester as defined in RATS Architecture.
+An entity is the same as the Attester Target Environment defined in RATS Architecture.
 
-An entity corresponds to a "system component" as defined in the Internet Security Glossary {{RFC4949}}.
+An entity also corresponds to a "system component" as defined in the Internet Security Glossary {{RFC4949}}.
 That glossary also defines "entity" and "system entity" as something that may be a person or organization as well as a system component.
 Here "entity" never refers to a person or organization.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-eat.md
```diff
@@ -236,10 +236,44 @@ The notion of attestation here is large and may include, but is not limited to t
...
 In particular it is a format that can be used for Attestation Evidence or Attestation Results as defined in the RATS architecture.
 
+## Entity Overview
+
+The document uses the term "entity" to refer to the target of the attestation token.
+The claims defined in this document are claims about an entity.
+
+An entity is an implementation in hardware, software or both.
+
+An entity roughly corresponds to a "system component" as defined in the Internet Security Glossary {{RFC4949}}.
```
henkbirkholz, line 246: As a read I do know what to unpack from "roughly" here. It is the main semantic anchor of the document.

henkbirkholz, line 246:
```suggestion
An entity corresponds to a "system component" as defined in the Internet Security Glossary {{RFC4949}}.
```

henkbirkholz: If that feels like it is not enough, please add text that specifically elaborates on the difference.

draft-ietf-rats-eat.md
```diff
@@ -236,10 +236,44 @@ The notion of attestation here is large and may include, but is not limited to t
...
+That glossary also defines "entity" and "system entity" as something that is a person or organization as well as a system component.
+Here "entity" never refers to a person or organization.
+
+An entity is never a server or a service.
+
+An entity roughly corresponds to an Attester as defined in RATS Architecture.
```
henkbirkholz, line 252: Same here

henkbirkholz, line 252:
```suggestion
An entity corresponds to an Attester as defined in RATS Architecture.
```

henkbirkholz (changes requested):

PR #150, opened 2022-01-13
Remove boot-seed and uptime claims

laurencelundblade:
Non-critical claims; make the document a little shorter.


gmandyam:
We have identified cases for uptime.  Moreover I could not identify a request to remove these 2 claims based on the list feedback, nor the issues list.  Recommend withdrawing this PR.

laurencelundblade:
Hannes and others have said the document is too long and that it doesn't need to define so many claims. These are the two claims I thought were the first on the list to go.

I'm personally neutral on whether they stay or go. I'll leave this PR open for a while and withdraw it if no one else wants them removed.

gmandyam:
@laurencelundblade 

It seems that an alternative candidate for claim removal are the DLOA claim - which is not clear that anyone will make use of it outside of GP, which seems to already have this information covered independent of EAT.  Security level may also be a candidate for removal given that it does not provide additional security assurances to a relying party - it is self-attested claim.   Also, removal of the current Section 4 (if we are interested in shortening the doc) may be in order as it is not clear that a discussion on unsigned JWT's should be addressed in an attestation document.

henkbirkholz:
Unprotected JWT Claims Sets are now at least covered in Appendix A of the UCCS I-D: https://www.ietf.org/rfcdiff?url2=draft-ietf-rats-uccs-02.txt

laurencelundblade:
Let's keep issues separate and independent.  If you want to remove DLOA's and security level please file separate issues for them.

We already have issues filed for the UCCS stuff.

gmandyam:
>Let's keep issues separate and independent. If you want to remove DLOA's and security level please file separate issues for them.

Actually I was responding to your rationale on selecting these 2 claims:

>These are the two claims I thought were the first on the list to go.

This seems to be based on arbitrary criteria, so I suggested two others that could be "first on the list to go".

PR #158, opened 2022-01-28
IANA registrations for the application/eat+cbor media type

### LINKED ISSUE #157: a media type for EAT
It looks like we are missing an explicit media type for EAT.
 
One important thing is that we make the media type so that it can optionally carry the EAT profile.  This allows HTTP/CoAP APIs to dispatch to the right decoder without having to snoop into the payload.
 
Something like `application/eat+cbor` for the base thing and an optional `profile` parameter that mirrors the profile claim value in the EAT. E.g.: `application/eat+cbor; profile=1.3.6.1.4.1.123.456`.

See also the RATS ML thread starting at https://mailarchive.ietf.org/arch/msg/rats/VAYmwacEeu-vIVjotbVqmOavQpk/

thomas-fossati:
Fixes #157

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1934,7 +1934,46 @@ specification reference.
 | Tag    | Data Items     | Semantics                   |
 | TBD602 | array          | Detached EAT Bundle {{DEB}} |
 
+## Media Type Registration
 
+IANA is requested to register the "application/eat+cbor" media type in the
+"Media Types" registry {{!IANA.media-types}} in the manner described in
+{{!RFC6838}}, which can be used to indicate that the content is a EAT object:
+
+* Type name: `application`
+* Subtype name: `eat+cbor`
+* Required parameters: n/a
+* Optional parameters: "profile" (the EAT profile claim in string format.  OIDs
+  MUST use the dotted-decimal notation.)
+* Encoding considerations: binary
+* Security considerations: See the Security Considerations of <cref>RFCTHIS</cref>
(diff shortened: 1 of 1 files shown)


gmandyam:
a) This PR lacks detail as to when to use the proposed media type in place of application/CWT.

b) This PR does not state whether eat+cbor excludes COSE/CWT encapsulation.

c) This PR lacks a corresponding section for security considerations, particularly in the presence of content-filtering proxies.  EAT was designed to provide a self-protecting payload, which could therefore be transported without any corresponding x-port security (e.g. TLS/DTLS).  However an intermediary could choose to drop an unrecognized content type if it can peek into the headers, and many existing proxies will.  The original intent of the editors was to avoid defining new media types to allow for seamless transition with existing proxies/gateways, without having to compel the use of a specific type of transport security.  Contrast this to protocols such as HTTP 2.0, which had to essentially mandate the use of TLS 1.2 (https://datatracker.ietf.org/doc/html/rfc7540#section-9.2) in practice (https://daniel.haxx.se/blog/2015/03/06/tls-in-http2/) to ensure in part that outdated proxies would not drop the traffic.  What is the security guidance on the use of new media types?  

I don't think this PR is ready for inclusion in the EAT spec, and I think this topic would be better handled more comprehensively in a separate draft that provides for all attestation media types, along with sufficient guidance for implementors, and associated security considerations.  Moreover, this topic seems broader than just attestation.  Many data types that appear in a protocol that involves attestation conveyance may benefit from a more expressive CWT/JWT media type.

PR #159, opened 2022-01-29
Clarifications to the submods section

laurencelundblade:
This is based on comments from Hannes.


laurencelundblade:
Something is wrong with this PR. It's not picking up the last push.

PR #162, opened 2022-01-30
Clarify the security level claimed 

laurencelundblade:
From mailing list comment by MCR.

Laurence Lundblade <lgl@island-resort.com> wrote:
- It is claiming the lowest level possible (as a warning)

I think that this is the key part that makes it useful, even when self-asserted.

It is saying, "I am *least* THIS vulnerable",
         not "I am at *most* THIS secure".


gmandyam:
"The security level claimed MUST be for the weakest point in the entity, not the strongest.  For example, if attestation key is protected by hardware, but the rest of the attester is in a TEE, the claim must be for secure-restriced.".  

In my opinion this does not fully capture the original intent, which was to convey a security level whose value could be determined independent from the entity manufacturer.  Therefore, in the example above, a manufacturer of an entity should not be able to claim a TEE-level of security assurance if this level of assurance has not been independently verified (e.g. by a cert house).     I would recommend adding text to clarify the difference between a security level that has been self-asserted versus a security level that has been independently verified.

I suggest the following:

"The security level claimed MUST be for the weakest point in the entity, not the strongest.  The weakest point of the entity could be determined by the attester itself, in which case it is self-determined.  The weakest point may also be determined by independent analysis (e.g. via a 3rd-party security certification program), in which case the security level is not self-asserted and can be confirmed separately from the attestation (e.g. NIST's Cryptographic Module Validation Program online lookup - see https://csrc.nist.gov/projects/cryptographic-module-validation-program/validated-modules/search).  

For example, in the case of self-assertion if the attestation key is protected by hardware, but the rest of the attester is in a TEE, the claim must be for secure-restricted.  However, if the security level is not self-determined then the attester should only claim the security level at which the entity was verified by an independent 3rd-party.  In this example, if the manufacturer only obtained a security certification for the entity that corresponded to 'restricted' level, then that is the level at which the claim should be set even though the entity contains a TEE".

PR #196, opened 2022-06-10
UEID length and type clarifications

laurencelundblade:
Addresses comments from MCR and Elliot.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -534,9 +534,15 @@ and to have an alternative that doesn't require paying a registration fee.
 
 Creation of new types requires a Standards Action {{RFC8126}}.
 
-UEIDs are variable length. All implementations MUST be able to receive
-UEIDs that are 33 bytes long (1 type byte and 256 bits).
-No UEID longer than 33 bytes SHOULD be sent.
+UEIDS are variable length to accommodate the types defined here and new types that may be defined in the future.
+
+All implementations MUST be able to receive UEIDs up to 33 bytes long.
+33 bytes is the longest defined in this document and gives necessary entropy for probabilistic uniqueness.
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-eat.md
```diff
@@ -534,9 +534,15 @@ and to have an alternative that doesn't require paying a registration fee.
...
-UEIDs that are 33 bytes long (1 type byte and 256 bits).
-No UEID longer than 33 bytes SHOULD be sent.
+UEIDS are variable length to accommodate the types defined here and new types that may be defined in the future.
+
+All implementations MUST be able to receive UEIDs up to 33 bytes long.
+33 bytes is the longest defined in this document and gives necessary entropy for probabilistic uniqueness.
+See {{UEID-Design}}.
+
+UEIDs SHOULD NOT be longer than 33 bytes.
+If they are longer, there is no guarantee that a receiver will be able to accept them.
```
mcr, line 544: Again, SHOULD. Is there an exception then?

draft-ietf-rats-eat.md
```diff
@@ -564,6 +570,9 @@ this are:
   may find they can optimize their process by switching from type 0x01
   to type 0x02 or vice versa.
 
+The type byte is needed to distinguish UEIDs of different types that by chance have the same identifier value, but do not identify the same entity.
+The type byte MUST be treated as part of the opaque UEID and MUST not be used to make use of the internal structure of the UEID.
```
mcr, line 574: What do the EUI and IMEI types bring to the table?  
Why not just have a 32-byte RAND UEID and be done with it?

laurencelundblade: EUI and IMEI are much shorter, they may be ready-to-hand and and a random number generator might not be.

PR #197, opened 2022-06-10
Address comments on SUEIDs

laurencelundblade:
Address Eliot and MCR's comments on SUEIDs.

This is just clarifications. No actual change to the definition of an SUEID.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -573,17 +573,23 @@ A Device Identifier URN is registered for UEIDs. See {{registerueidurn}}.
 
 ### Semi-permanent UEIDs (SUEIDs)
 
-An SEUID is of the same format as a UEID, but it MAY change to a different value on device life-cycle events.
-Examples of these events are change of ownership, factory reset and on-boarding into an IoT device management system.
+An SUEID has the same format, characteristics and requirements as a UEID, but MAY change to a different value on entity life-cycle events.
 An entity MAY have both a UEID and SUEIDs, neither, one or the other.
 
+Examples of life-cycle events are change of ownership, factory reset and on-boarding into an IoT device management system.
+It is beyond the scope of this document to specify particular types of SUEIDs and the life-cycle events that trigger their change.
+An EAT profile MAY provide this specification.
+
 There MAY be multiple SUEIDs.
-Each one has a text string label the purpose of which is to distinguish it from others in the token.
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-eat.md
```diff
@@ -573,17 +573,23 @@ A Device Identifier URN is registered for UEIDs. See {{registerueidurn}}.
...
 ### Semi-permanent UEIDs (SUEIDs)
 
-An SEUID is of the same format as a UEID, but it MAY change to a different value on device life-cycle events.
-Examples of these events are change of ownership, factory reset and on-boarding into an IoT device management system.
+An SUEID has the same format, characteristics and requirements as a UEID, but MAY change to a different value on entity life-cycle events.
 An entity MAY have both a UEID and SUEIDs, neither, one or the other.
 
+Examples of life-cycle events are change of ownership, factory reset and on-boarding into an IoT device management system.
+It is beyond the scope of this document to specify particular types of SUEIDs and the life-cycle events that trigger their change.
+An EAT profile MAY provide this specification.
```
mcr, line 581: Since EUI and IMEI labels come from a registry, they can't change during a life-cycle event.
It seems that really only the RAND type makes any sense for the SUEID.

draft-ietf-rats-eat.md
```diff
@@ -573,17 +573,23 @@ A Device Identifier URN is registered for UEIDs. See {{registerueidurn}}.
...
 There MAY be multiple SUEIDs.
-Each one has a text string label the purpose of which is to distinguish it from others in the token.
+Each has a text string label the purpose of which is to distinguish it from others.
 The label MAY name the purpose, application or type of the SUEID.
-Typically, there will be few SUEDs so there is no need for a formal labeling mechanism like a registry.
-The EAT profile MAY describe how SUEIDs should be labeled.
-If there is only one SUEID, the claim remains a map and there still must be a label.
-For example, the label for the SUEID used by FIDO Onboarding Protocol could simply be "FDO".
+For example, the label for the SUEID used by FIDO Onboarding Protocol could be "FDO".
+It is beyond the scope of this document to specify any SUEID labeling schemes.
```
mcr, line 587: So, it's okay for me to use FDO to mean something else?

laurencelundblade: Yes.

It doesn't quite say it explicit, but if you are really concerned about collisions with these labels in a particularly use case then define a profile which says how they are allocated.

PR #210, opened 2022-06-18
Clarify how a profile can re-define an EAT claim

gmandyam:
See https://mailarchive.ietf.org/arch/msg/rats/zubR3clEkOwh68bPvwbD8NPwY2k/ (ioT Directorate review):

"This statement in particular is quite disturbing.

   In some cases CDDL may be created that replaces CDDL in this or other
   document to express some profile requirements.

Not only is this counter-cultural, but it would require an Updates:
header on any such profile, and would further just be plain out
confusing.

In short, the profile mechanism is harmful to the very concept of
interoperability"

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1494,8 +1494,7 @@ A profile can apply to Attestation Evidence or to Attestation Results or both.
 
 A profile document doesn't have to be in any particular format. It may be simple text, something more formal or a combination.
 
-In some cases it may be desirable for CDDL to be created that replaces CDDL in this or other document to express some profile requirements.
-For example, to require the altitude data item in the location claim, CDDL can be written that replicates the location claim with the altitude no longer optional.  In that case, the profile MUST define a new claim in the place of the original EAT claim and leverage the appropriate IANA registries to point to assign an appropriate claim identifier and pointer to the profile specification.
+A profile may define, and possibly register, one or more new claims if needed. A profile may also reuse one or more already defined claims, either as-is or with values constrained to a subset or subrange.
 
 ## List of Profile Issues
 
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1494,8 +1494,8 @@ A profile can apply to Attestation Evidence or to Attestation Results or both.
 
 A profile document doesn't have to be in any particular format. It may be simple text, something more formal or a combination.
 
-In some cases CDDL may be created that replaces CDDL in this or other document to express some profile requirements.
-For example, to require the altitude data item in the location claim, CDDL can be written that replicates the location claim with the altitude no longer optional.
+In some cases it may be desirable for CDDL to be created that replaces CDDL in this or other document to express some profile requirements.
+For example, to require the altitude data item in the location claim, CDDL can be written that replicates the location claim with the altitude no longer optional.  In that case, the profile MUST define a new claim in the place of the original EAT claim and leverage the appropriate IANA registries to point to assign an appropriate claim identifier and pointer to the profile specification.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-eat.md
```diff
@@ -1494,8 +1494,8 @@ A profile can apply to Attestation Evidence or to Attestation Results or both.
 
 A profile document doesn't have to be in any particular format. It may be simple text, something more formal or a combination.
 
-In some cases CDDL may be created that replaces CDDL in this or other document to express some profile requirements.
-For example, to require the altitude data item in the location claim, CDDL can be written that replicates the location claim with the altitude no longer optional.
+In some cases it may be desirable for CDDL to be created that replaces CDDL in this or other document to express some profile requirements.
+For example, to require the altitude data item in the location claim, CDDL can be written that replicates the location claim with the altitude no longer optional.  In that case, the profile MUST define a new claim in the place of the original EAT claim and leverage the appropriate IANA registries to point to assign an appropriate claim identifier and pointer to the profile specification.
```
thomas-fossati, line 1498: If this means that we'd need to renumber the [nonce](https://datatracker.ietf.org/doc/html/draft-tschofenig-rats-psa-token#section-3.1.1) and the the [instance ID](https://datatracker.ietf.org/doc/html/draft-tschofenig-rats-psa-token#section-3.2.1) only because they express a constraint over the EAT type I am definitely opposed to this change.

thomas-fossati (changes requested):

draft-ietf-rats-eat.md
gmandyam: You are welcome to take your concerns to the Working Group and the IETF IoT Directorate and try to build consensus.  IANA registration requires a stable specification, and the nonce claim has been registered as per the current definition in the EAT document.

draft-ietf-rats-eat.md
```diff
@@ -1494,8 +1494,8 @@ A profile can apply to Attestation Evidence or to Attestation Results or both.
 
 A profile document doesn't have to be in any particular format. It may be simple text, something more formal or a combination.
 
-In some cases CDDL may be created that replaces CDDL in this or other document to express some profile requirements.
-For example, to require the altitude data item in the location claim, CDDL can be written that replicates the location claim with the altitude no longer optional.
+In some cases it may be desirable for CDDL to be created that replaces CDDL in this or other document to express some profile requirements.
```
thomas-fossati, line 1497:
```suggestion
In some cases it may be desirable for CDDL to be created that constrains CDDL in this or other document to express some profile requirements.
```

draft-ietf-rats-eat.md
thomas-fossati, line 1498:
```suggestion
```

draft-ietf-rats-eat.md
henkbirkholz: I like the focus on actual "profiling", which literally means "constraining" to a subset. Thank you Thomas, for that clarification. I am seconding this suggestion.

draft-ietf-rats-eat.md
henkbirkholz: As in the above, I 100% agree with the increase of a constraint in a profile. 2nding this suggestion.

henkbirkholz:
I would like to try to read Eliot's comment here.

The way profiling is phrased does not work. We have to:

1.) define the hard issue with current profiling
2.) agree on what profiling in general means

Without theses two actions done on the list, in the open, with consensus, any tinkering on this issues is moot.

gmandyam:
Made another attempt to clarify profile usage of existing claims.

thomas-fossati (approved):

PR #214, opened 2022-06-19
Clarify which epoch time is used for iat

laurencelundblade:
Per comment from Eliot

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1279,10 +1279,12 @@ indicate their age is older than the "iat" timestamp.
 
 CWT allows the use floating-point for this claim. EAT disallows
 the use of floating-point. An EAT token MUST NOT contain an iat claim in
-float-point format. Any recipient of a token with a floating-point
-format iat claim MUST consider it an error.  A 64-bit integer
-representation of epoch time can represent a range of +/- 500 billion
-years, so the only point of a floating-point timestamp is to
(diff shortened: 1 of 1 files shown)


henkbirkholz (changes requested):
Referring indirectly to:

> [TIME_T]   The Open Group, "The Open Group Base Specifications",
              [Section 4.16](https://datatracker.ietf.org/doc/html/rfc8949#section-4.16), 'Seconds Since the Epoch', Issue 7, 2018
              Edition, IEEE Std 1003.1, 2018,
              <[https://pubs.opengroup.org/onlinepubs/9699919799/](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_16)
              [basedefs/V1_chap04.html#tag_04_16](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_16)>.

does not really help here. "Epoch based time" only makes sense in context. You mean "POSIX, IEEE Std 1003, Seconds since Epoch"? Say that. Every instantiation of #6.1002 would have it's individual epoch, initiating its own small timescale for the duration, starting at 0.

I's assume the at you want to reference and in text elaborate on the most recent IEEE posix document. The current change does not add clarity, I think.

laurencelundblade:
The use of "epoch-based" here is to reference the section in RFC 8949.  I don't think every time you use CBOR epoch time you have to reference POSIX.

PR #216, opened 2022-06-22
Add standard constrained device profile

laurencelundblade:


draft-ietf-rats-eat.md
```diff
@@ -1623,6 +1623,33 @@ The profile should specify which formats are allowed for the manifests and softw
...
+| CBOR Serialization | Only preferred serialization is allowed |
+| COSE Protection | Only COSE_Sign1 format is used |
+| Algorithms | Receiver must accept ES 256, ES384 and ES 512. Sender must send one of these |
+| DEB Usage | DEB may not be sent with this profile |
+| Verification Key Identification | Either the COSE kid or the UEID MUST be used to identify the verication key. If both are present, the kid takes precedence |
+| Endorsements | This profile contains no endorsement identifier |
+| Nonce | A new single unique nonce must be used for every token request |
+| Claims | No requirement is made on the presence or absence of claims. The general EAT rules apply -- the nonce must be present and the receiver must not error out on any claims it doesn't understand|
+
+Stricly speaking, slight modifications such use of a different means of key identification are a divergence from this claim and should use a different profile identifier.
```
thomas-fossati, line 1647:
```suggestion
Strictly speaking, slight modifications such use of a different means of key identification are a divergence from this profile and MUST use a different profile identifier.
```

draft-ietf-rats-eat.md
```diff
@@ -1623,6 +1623,33 @@ The profile should specify which formats are allowed for the manifests and softw
 The profile may also go on to say which parts and options of these formats are used, allowed and prohibited.
 
 
+## The Constrained Device Standard Profile
+
+It is anticipated that there will be many profiles defined for EAT for many different use cases.
+This section standardizes one profile that is good for many constrained device use cases.
+
+The identifier for this profile is "https://tools.ietf.org/RFC-TBD"
```
thomas-fossati, line 1631: We should ask if there is any prior art on allocating URIs under the IETF.org domain name.

draft-ietf-rats-eat.md
```diff
@@ -1623,6 +1623,33 @@ The profile should specify which formats are allowed for the manifests and softw
...
+
+
+| Issue | Profile Definition |
+| CBOR/JSON | CBOR only |
+| CBOR Encoding | Only definite length maps and arrays are allowed |
+| CBOR Encoding | Only definite length strings are allowed |
+| CBOR Serialization | Only preferred serialization is allowed |
+| COSE Protection | Only COSE_Sign1 format is used |
+| Algorithms | Receiver must accept ES 256, ES384 and ES 512. Sender must send one of these |
+| DEB Usage | DEB may not be sent with this profile |
```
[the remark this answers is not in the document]
laurencelundblade: I kind of prefer your wording here, but I've had some comments that say it should be more of the MUST/SHOULD/MAY form.

draft-ietf-rats-eat.md
laurencelundblade: Not sure who to ask. I don't think it is IANA.

Note that I'm just using the URL for the EAT document itself. There's not really any allocation required, but it does need to be set to the right number. 

Maybe we just need a note to the RFC editor?

thomas-fossati: Looking at [the RFC Editor archival policies](https://www.rfc-editor.org/rfc/rfc8153.html#section-2.3), it seems that something like `https://www.rfc-editor.org/rfc/rfcXXXX` would be the stable URI we would be looking for.  (Note that tools.ietf.org is being [decommissioned](https://www.ietf.org/blog/finalizing-ietf-tools-transition/) as we speak.)

An alternative is to grab an OID from somewhere, though I wouldn't know under which arc as there doesn't seem to be one for IETF.

thomas-fossati: however, a more fundamental question is: do we need an identifier or can we leave it "anonymous"?  The fact is that without defining the acceptable claims-set I suspect no one will use it as-is, but only by normatively referencing the section / table and slapping their claims-set on.  That combination is what will be given a profile identifier.

laurencelundblade: OK. Fixed the URI. I like the idea that it is there as an example and that you can find the document by the URI.

Filed an issue for a JSON profile. Don't have time to do it for the cutoff date.

I am merging this so it can be in the next draft publication.

Thanks for the help here.

PR #223, opened 2022-06-29
Add SPDX and CycloneDX to manifests claim

laurencelundblade:


gmandyam:
Do not understand sending XML in a JS object.  XML would have to be cast to JSON first.  There have been attempts to standardize this casting in the past, e.g. see https://datatracker.ietf.org/doc/draft-rsalz-jsonx/.  Would recommend removing the CycloneDX XML option alltogether.

laurencelundblade:
Well, um we mix formats regularly. There's no CBOR format for CycloneDX. That doesn't/shouldn't stop putting a JSON or XML format CycloneDX into a CBOR token. 

 There's a lot more formats for SPDX and CycloneDX. I just registered the ones for which there were MIME types.

gmandyam:
>Well, um we mix formats regularly. There's no CBOR format for CycloneDX. That doesn't/shouldn't stop putting a JSON or XML format CycloneDX into a CBOR token.

That isn't really relevant to the point I was making.  Casting XML to JSON (or CBOR) is not automatic just because we define some CDDL in the EAT spec.  XML must be first converted to a JS or CBOR object, and there are multiple ways of doing this.  

It is certainly possible to serialize the XML before including it in a token, but the CDDL as provided in the PR does not seem to address this.  

At very least, we should hold off on incorporating this PR until we get someone with more in-depth familiarity with SPDX and CycloneDX  to review.

laurencelundblade:
Agree with holding off the PR if you want. Presumably that means it won't go into the draft we're about to publish.

The EAT CDDL specifically defines the  plug-in to the manifests claim socket for the XML and JSON encodings of CycloneDX be text so that the actual XML and JSON can be put in the EAT, whether the EAT is CBOR or JSON.. The manifests is specifically constructed to allow manifests that are encoded differently than the surrounding token. This is a must because the manifests encoding format will often not match the EAT encoding format. For example a CBOR-encoded SUIT manifest or a CBOR CoSWID have to be able to go into a JSON-encoded token because there is no JSON format for SUIT or CoSWID.

gmandyam:
>The EAT CDDL specifically defines the plug-in to the manifests claim socket for the XML and JSON encodings of CycloneDX be text so that the actual XML and JSON can be put in the EAT, whether the EAT is CBOR or JSON..

Yes, this works.  But this requirement is explicitly spelled out in the current SUIT manifests text:  https://datatracker.ietf.org/doc/html/draft-ietf-rats-eat#section-4.2.16.

For instance, in the context of SUIT manifests the current EAT spec specifically states:  "In CBOR-encoded EATs the manifest, whatever format it is, MUST be placed in a byte string."  Is that the intention for the SBOM formats as described in this PR?  If so, how would an independent reader determine this?

laurencelundblade:
Ahhh... that text was wrong. See #234. I just removed it and rely on the CDDL.

I think it all works now.

We're agreed that it is OK for the encoding of the manifest to be different from the encoding of the token, right? Seems like it has to be that way because there are too many manifest formats and very few have both JSON and CBOR encodings.

gmandyam:
>We're agreed that it is OK for the encoding of the manifest to be different from the encoding of the token, right? Seems like it has to be that way because there are too many manifest formats and very few have both JSON and CBOR encodings.

Agreed - seems to be no other way to do it.  The proposed text is now consistent with the SUIT text and can be merged.

PR #229, opened 2022-07-05
Update definition of sec level 'restricted'

gmandyam:
Current reference pointing to FIDO AROE is misleading - AROE is a definition associated with authenticators and is not applicable as a MUST requirement for an entity security level.  Moreover, EAT is not supported in any FIDO authentication standards.


laurencelundblade:
Hi Giri,

AROE is not authentication-specific. Look through sections 3 and 4 of https://fidoalliance.org/specs/fido-security-requirements-v1.0-fd-20170524/fido-authenticator-allowed-restricted-operating-environments-list_20170524.html to confirm.

It doesn't matter to the definition of security-level that FIDO won't use EAT.  

The change you propose would substantially weaken the definition of security level. A lot of useful and thought and work went into the definition of an ROE and it is helpful to the definition of security-level.

gmandyam:
FIDO documents should be interpreted in the FIDO context.  here are two broad categories of FIDO specifications:  authenticator (UAF, U2F, FIDO 2.0) and IoT (FIDO Device Onboard).  The definition of AROE is not applicable to FDO and there is no mention of AROE in the FDO security certification requirements.  Therefore it is narrowly scoped to authenticators.  In fact, it is not applicable at all to FDO devices, which are the only FIDO-compliant devices that produce an EAT attestation.  

Moreover, the current text only refers to the AROE document Section 4, and this document only specifies AROE's as per Section 3.  This is problematic, as it excludes Secure MCU definition as specified in SESIP - see https://globalplatform.org/wp-content/uploads/2021/06/GP_SESIP_Profile_Secure_MCU_MPU_v0.0.0.7_clean.pdf.  It seems rather strange to point to Sec. 4 of the AROE document but advise the EAT audience to ignore Section 3.  I don't see much benefit in trying to stretch this document beyond what it was intended for - defining a restricted operating environment for authenticators.

If you don't like my suggested text, then Section 2 of the SESIP document referenced above could be a better reference.  But the AROE document simply doesn't work.

>It doesn't matter to the definition of security-level that FIDO won't use EAT.

The above statement is inaccurate. FDO does use EAT.  See https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#device-attestation.

henkbirkholz:
Just as a note:

FDO == FIDO Device Onboard
AROE == Allowed Restricted Operating Environments
SESIP == Security Evaluation Standard for IoT Platforms (from Global Platform, one of their "Endorsement types" and I think a current EAT sec-level?)

laurencelundblade:
I've added text that clarifies that only section 4 is referenced. Section 3 does not apply.

It's OK to reference only part of some document, right? We do that all the time. For example CDDL references regexp from some other massive document. 

It doesn't matter to _this definition_ and the EAT standard whether any FIDO protocol uses security level or doesn't use security level, uses EAT or doesn't use EAT. This definition only makes normative reference to section 4 of the AROE document, nothing more.

I would like this PR merged as is for the next draft publication.

gmandyam:
>It's OK to reference only part of some document, right? We do that all the time. For example CDDL references regexp from some other massive document.

I don't think so.  Context matters. The AROE document itself nowhere states that Section 4 is not related to Sections 3 or 2.  In fact, it can't be because the term AROE is never actually defined in Section 4. Even the acronym isn't expanded, so the user has to refer to Section 3.  And Section 3 is clearly constrained to the table of AROE's that are applicable to FIDO certification.  

There are better references.  Why are we not using them?  I mentioned the SESIP secure MCU document, which unfortunately is excluded from the AROE document alltogether.  Another good reference is https://www.commoncriteriaportal.org/files/ppfiles/anssi-profil_PP-2014_01.pdf.  I think replacing the AROE reference with these two references for Secure MCU and TEE respectively is a vast improvement over the existing text.

laurencelundblade:
Hi Giri,

I went a looked over the SESIP Profile. Then I looked at SESIP itself. This is one hundred plus pages of definition between the two documents, and a lot more pages if you follow all the normative references. It is also evaluation program, maybe even a certification program. It seems kind of like a light version of Common Criteria. 

If this were to appear in an EAT, I think it should be via a DLOA. It's not really in the spirit of the security-level claim, which should not require any 3rd party evaluation and expense. It seems much too large and heavy. 

You are right about the term "AROE" used in section 4. Section 4 doesn't really stand on its own because of the use of "AROE". I'm sorry I missed that before.

However, If you take the text in section 4 and replace the term "AROE", with "entity" it does work. There is no other reference to anything else in FIDO. It stands on its own like many subsections that have been referenced in lots of documents. I tried it to see: https://github.com/laurencelundblade/security-level/blob/master/security-level.md

It would be really awkward to say something in EAT like "must meet the criteria in section 4 where the term 'AROE' is replaced with the term 'entity'."

So I still think AROE definition suits the purpose, but the wording, the use of the term "AROE" kind of messes up the ability to reference it.

I am still thinking about what we might do... more later today.

gmandyam:
>If this were to appear in an EAT, I think it should be via a DLOA. It's not really in the spirit of the security-level claim, which should not require any 3rd party evaluation and expense. It seems much too large and heavy.

I don't see it that way.  DLOA can only be provided when a certification is achieved.  That doesn't mean that the entity cannot satisfy a set of SFR's and map them to a sec level, regardless of whether a DLOA is applicable or not.  The SESIP and TEE PP documents I cited list a set of SFR's that the entity can claim to meet based on the declared sec level.

That being said, another possible alternative to the AROE document is NIST 800-164:  https://csrc.nist.gov/publications/detail/sp/800-164/draft.  In this doc, the characteristics of a HW RoT are described and clearly distinguished from a TPM - so in that regards it may be suitable for Restricted level.  That being said, this document is focused on mobile devices so an independent reader may conclude that it is not applicable to IoT.

The ETSI cybersec requirements for IoT could however serve the purpose of defining requirements for non-mobile devices (see Sec. 5 of https://www.etsi.org/deliver/etsi_en/303600_303699/303645/02.01.01_60/en_303645v020101p.pdf).  So between the two documents most entities that meet 'restricted' level would be covered.

laurencelundblade:
security-level is being removed

laurencelundblade:
security-level is being removed

laurencelundblade:
security-level is being removed

PR #247, opened 2022-08-01
Address some of the high-level intro-type comments

laurencelundblade:
This addresses some of the framing comments from Hilarie and Eliot.


laurencelundblade:
This is replaced by #303 that focuses just on Appendix C and #279 that focuses just on the abstract and some of the other in-process PRs focused just on the intro sections

PR #248, opened 2022-08-02
Change some terms to lowercase

carl-wallace:
Change these terms to lowercase for readability per comment from Eliot Lear on June 5:

- Relying Party
- Attester
- Verifier
- Verifier's Policy
- Evidence
- Attestation Results
- Reference Values
- Endorsement
- Endorsement Identification
- Appraisal Policy
- Verification Key Identification
- Policy

Change Attester Target Environment to attesting environment to align with current architecture document.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -264,7 +264,7 @@ Here are a few examples of claims:
 EAT also supports nesting of sets of claims and EAT tokens for use with complex composite devices.
 
 This document uses the terminology and main operational model defined in [RATS.architecture].
-In particular, it can be used for Evidence and Attestation Results.
+In particular, it can be used for evidence and attestation results.
 
 ## Entity Overview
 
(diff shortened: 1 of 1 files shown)


laurencelundblade:
Hi Carl,

I put some time into this and investigated this a bit with the RFC Editor and the RATS Architecture authors and decided to go with the caps because that seemed the easiest. It is the convention adopted by the RATS WG. It seemed like it would be more trouble to argue against or do different.

I kind of prefer the non-capitalization now that I see it, but what I prefer the most is less work and timely completion.

So, I'm fine with the change, but if the WG comes back and argues to return to capitalization, that is yours to resolve.

carl-wallace:
Works for me. 

 

Date: Wednesday, August 3, 2022 at 1:00 PM
Cc: Carl Wallace ***@***.***>, Author ***@***.***>
Subject: Re: [ietf-rats-wg/eat] Change some terms to lowercase (PR #248)

 

Hi Carl,

I put some time into this and investigated this a bit with the RFC Editor and the RATS Architecture authors and decided to go with the caps because that seemed the easiest. It is the convention adopted by the RATS WG. It seemed like it would be more trouble to argue against or do different.

I kind of prefer the non-capitalization now that I see it, but what I prefer the most is less work and timely completion.

So, I'm fine with the change, but if the WG comes back and argues to return to capitalization, that is yours to resolve.

Issue #246, opened 2022-08-01
Clarify algorithm claim in DEB digest


gmandyam:
It is not clear from the existing text that the algorithm can only be drawn from the COSE/JOSE algorithms registry, except for in the profiles section:  see https://datatracker.ietf.org/doc/html/draft-ietf-rats-eat#section-6.2.7.  Even in the context of profiles, only the COSE algorithm registry is mentioned and not the JOSE registry (https://www.iana.org/assignments/jose/jose.xhtml).  As a result, since the algorithm is not specifically to be drawn from one of the existing registries (except for CWT-based profiles of EAT), then the worst-case assumption is that the algorithm could be arbitrary (and arbitrarily weak).  

If the algorithmic identifier from either the COSE or JOSE registries are to be used, then it would be desirable to explicitly define an algorithm claim which would be consistent data types in the CWT/JWT headers respectively.

laurencelundblade:
Fixed in #250

PR #250, opened 2022-08-05
Detached digest algorithms are from COSE registry

laurencelundblade:


laurencelundblade:
Address #246

gmandyam:
JOSE also has its own registry:  https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms.  For JWT variants of EAT/DEB do we consider these identifiers instead?

laurencelundblade:
I excluded the JOSE registry intentionally because a) it has no hash functions registered and b) detached digests are not particularly bound to whether the token is JSON or CBOR.

gmandyam:
@laurencelundblade 

> excluded the JOSE registry intentionally because a) it has no hash functions registered

Don't understand the above.  The first 3 entries in the JOSE registry are HMAC algm.'s,.  Or is the assertion that MAC's are never suitable for the DEB digest?

laurencelundblade:
Yes, it needs a hash algorithm, not an HMAC.

gmandyam:
>Yes, it needs a hash algorithm, not an HMAC.

If that is the case, then the text needs to be clarified that only hash algm.'s (and not MAC's) are suitable candidates from the COSE algm. registry.  HMAC's as digests comes up in different places, and an independent reader/implementer may not know the difference.

gmandyam:
Satisfied with PR now.  Please merge (unless there are other reviewers).

PR #252, opened 2022-08-08
Eliminate use of DEB as acronym for detached EAT bundle.

carl-wallace:
Used DEAT where an abbreviation was necessary (i.e., structure names and the type identifier for JSON-Nested-Token).


laurencelundblade:
I think we could just use the word "bundle" instead of "DEAT". I think it works in all the places you used "DEAT".

carl-wallace:
That works. I wasn't crazy about DEAT either. Changed. Thanks.

PR #253, opened 2022-08-09
Remove security level claim

laurencelundblade:
Includes removal from all examples.


carl-wallace:
The Security Level claim registration should probably be removed. Else looks good.

PR #256, opened 2022-08-11
Some guidance on use of submodules

laurencelundblade:


carl-wallace (commented):
Is it worth adding a MUST or SHOULD that labels in results should match those in evidence?

laurencelundblade:
No, I don't think so. There is no requirement or expectation that claims in evidence will pass through to results. 

For example, it's OK for there to be a UEID in evidence and none in results. It's even OK for there be a different UEID in results than in evidence perhaps because of a privacy preservation feature in the verifier.

It's OK for the submods in evidence to be structured completely different from the submods in results.

But one thing that needs to be true is that a claim whether used in results or in evidence aligns with the specification and semantics of its definition. For example, don't put SW Version information in the UEID claim. (duh).

The other things is that the verifier must say what their output means. For example, if they include a UEID in results, they should say whether they verified it or not. If they have submods they should say what they are so the RP knows how to process them.

gmandyam:
"Likewise, verifier ..."  -> "Likewise, a verifier ...".

Submodules do not need to be pass through, but they can be indirectly passed hrough.  For instance, attestation evidence could be for separate submodules "TEE" and "Linux Kernel", but verifier can group both of these into a submodule  called "Apps Subsystem".  Would the following textual refinement cover this:

"should define and document the submodules and their associated semantics for the relying party"

gkostal:
If a profile designer has a set of claims associated with a scenario (e.g. arbitrary "runtime claims" for a running trusted execution environment), how should they choose whether to place under a top level claim name (e.g. "runtime-claims") or a submod (e.g. named "runtime-claims")?  Mechanically, both will work in JWT/CWT, so I'm looking for some guidance in the spec here.

laurencelundblade:
@gkostal How about this?  -- generally submodules should be used to represent true subsystems, not as a claim grouping or classifying facility. A subsystem typically has separate hardware or at least a separate memory space (e.g. a process). (This applies to evidence as well as results).

I don't think this has to be a hard rule though.

I'm not sure what "arbitrary runtime claims" are and why they would be distinct from any other claims about the TEE. For example, I wouldn't think the TEE has a "real UEID" and an "arbitrary runtime UEID".

The other thing is to think about what the RP needs to know and what they'll do with the claims they get.

gkostal:
> @gkostal How about this? -- generally submodules should be used to represent true subsystems, not as a claim grouping or classifying facility. A subsystem typically has separate hardware or at least a separate memory space (e.g. a process). (This applies to evidence as well as results).
> 
> I don't think this has to be a hard rule though.
> 
> I'm not sure what "arbitrary runtime claims" are and why they would be distinct from any other claims about the TEE. For example, I wouldn't think the TEE has a "real UEID" and an "arbitrary runtime UEID".
> 
> The other thing is to think about what the RP needs to know and what they'll do with the claims they get.

@laurencelundblade - yes, that certainly helps.  Although "true subsystems" still leaves a lot to the interpretation of the reader.  :-)  It's the subsequent text regarding "not a claim grouping or a classifying facility" and "separate hardware/memory" that helps me distinguish when to use/not-use.  Perhaps EAT guidance on how to handle classification or claim grouping outside of a "true subsystem" would be helpful also?

I agree with the thought "not a hard rule though".

**Regarding "arbitrary runtime claims"**, I'm talking about the data that the code running in a TEE can associate with the running TEE -- for example, by recording the hash of this data in the "reportdata" field in either the Intel SGX report or the AMD SEV-SNP report.  In the Azure Attestation Service, we support a model where this arbitrary data can be structured as a JSON object with the resultant JWT produced by the service containing this JSON object intact as the value of a top level claim named "x-ms-runtime".  We've spent many hours discussing whether this JSON object should be placed as a value in a top level claim or placed as a value of a named submod.  

A common use for runtime claims is to share a public key value for an asymmetric key that's created & managed within the TEE that can be used to establish a secure communication channel with the TEE.

See the section on TEE runtime claims in this slide deck for more details on what I called "arbitrary runtime claims":
* https://github.com/CCC-Attestation/meetings/blob/main/materials/GregKostal_EAT_in_MAA.pdf

thomas-fossati:
I like Laurence's _"[...] generally submodules should be used to represent true subsystems, not as a claim grouping or classifying facility."_, I think that's the level of guidance we need.

However, I'd strengthen it a bit more like this: _"Submodules should not be used as a claim grouping or classifying facility.  Instead, they should be used to represent true subsystems (e.g., `INSERT_HERE_TRUE_SUBSYSTEMS_EXAMPLES`)."_

Besides, the requirements on profile designers to document the semantics of their submodules should be stated more strongly, i.e., using RFC2119 words.

gkostal:
If the spec states _"Submodules should not be used as a claim grouping or classifying facility"_ then it certainly (IMO) should state how claim grouping and classifying should be handled.

thomas-fossati:
> If the spec states _"Submodules should not be used as a claim grouping or classifying facility"_ then it certainly (IMO) should state how claim grouping and classifying should be handled.

Maybe: _"To that purpose, it is RECOMMENDED to define a claim with a value of type map that includes all the attributes that need to be grouped together."_

laurencelundblade:
Agreed on some better language around submods. Will work on that.

Following is some more off-the-top-of-my-head description of a "grouping claim".  Other ideas? There's nothing in EAT to stop anyone from defining one of these now. I'm not sure we want to put this text in EAT now because of the high pressure to get EAT done soon and because it might be a bigger can of worms that we think.

A "grouping claim" is a claim with a label that is of type map. Its content is other claims. Its purpose is to group some claims together that are related. It can contain any claim including another grouping claim. It cannot contain a submods section.

No "grouping claims" are defined in this document. They can be defined in other documents, profiles, proprietary uses of EAT and such.

When they are defined, the semantics associated with them should be defined. They should say what the criteria for being in the group is. This criteria may be strict or lax.

When decoding an EAT and encountering an unknown grouping claim, the decoder should ignore it just like any unknown claim. This will result in all the claims in the grouping claim being ignored. (There is no means to mark a grouping claim as such to allow traversal of it).

thomas-fossati:
> Agreed on some better language around submods. Will work on that.
> 
> Following is some more off-the-top-of-my-head description of a "grouping claim". Other ideas? There's nothing in EAT to stop anyone from defining one of these now. I'm not sure we want to put this text in EAT now because of the high pressure to get EAT done soon and because it might be a bigger can of worms that we think.

+1

I also think the idea of grouping claims is an important one, but we don't need to articulate it fully in this version of the EAT spec.  As long as there is some short but clear guidance along the lines that we have already discussed I am more than OK.

> A "grouping claim" is a claim with a defined label that is of type map. Its content is other claims. Its purpose is to group some claims together that are related. It can contain any claim including another grouping claim.

Note that limiting to maps may be unnecessarily restrictive: array is also an option (see [PSA SW components](https://thomas-fossati.github.io/draft-psa-token/draft-tschofenig-rats-psa-token.html#section-4.4.1)).

> It cannot contain a submods section.

I am not sure why? (we can discuss this and the rest of the points below in the context of a different draft.)

laurencelundblade:
I think the current PR is ready to merge. This is just comments on the future groups work.


If you allow arrays, then the elements in side the array can't be claims because they don't have claims labels/keys/ids.  Individual claims like PSA SW components can be arrays, but they are an individual claim, not a group of claims.

We don't allow more than one submod section now. Groups shouldn't change that. I also think that submods in groups would also complicate the decoder implementation a lot.

PR #263, opened 2022-08-12
Minor edits from Section 7 to end of doc

carl-wallace:
- Added reference to 8949 to time-int discussion
- Changed "Claim Characteristics" section title to "Design Guidance for Defining New EAT Claims". Changed some must/should language to MUST/SHOULD.
- Changed label in measurement results claim registration to match "measres" label used in prose.
- Minor edits


laurencelundblade:
I think we might need to pull the "meares" label change into a separate PR so we can get all the examples, labels and CDDL for it too.

carl-wallace:
I restored swresults, but I think it may just be one spot that needs to be changed additionally (in A.1.6).

Issue #238, opened 2022-07-16
say all claims are optional


laurencelundblade:
This is implied by EAT being a CWT and inheriting rules from it, but it is probably good to say it explicitly. There used to be text for it.

laurencelundblade:
Fixed by #264

Issue #219, opened 2022-06-27
Should nonce be optional?


laurencelundblade:
Right now an EAT MUST have a nonce.  However RATS architecture section 10.1 and TUDA do not require a nonce, using synchronized clocks instead.

I think we should relax the nonce requirement, but state that if a nonce is not used another means for freshness MUST be used.

laurencelundblade:
Fixed by #264

PR #264, opened 2022-08-16
Claim optionality, nonces and freshness

laurencelundblade:
This PR started out with the idea that nonces are optional because there are other ways to do freshness. This resulted in review of all the nonce and freshness-related sections and a lot of changes due to incorrect stuff. 

This also address the all-claims-are optional issue.

This needs review and more cross-checking before merging.

Addresses: #238 and #219


gmandyam:
Existing replay discussion in Privacy section was requested by RATS arch. team and was reviewed by them. See https://github.com/ietf-rats-wg/architecture/issues/399

PR #271, opened 2022-08-22
Remove exact duplicate examples that got added in draft 13 by accident

laurencelundblade:
Document is back under 100 pages.


laurencelundblade:
Really close

Issue #237, opened 2022-07-10
Normative references to RATS Architecture Security Considerations 


laurencelundblade:
There should be a strong normative reference to the extensive security considerations in RATS architecture, all of which apply.

gmandyam:
Needed for security considerations section

PR #274, opened 2022-08-24
Sec cons update with RATS arch. reference

gmandyam:
To address https://github.com/ietf-rats-wg/eat/issues/237


laurencelundblade:
Looks good to me.

PR #279, opened 2022-08-31
Finalize abstract

laurencelundblade:
I'd like to use this thread to finalize the abstract. I won't merge it until us authors are agreed on the final abstract.

This change addresses a comment by Eliot that EAT is more than just extensions to JWT/CWT.

Carl, I looked at your proposed changes #259 and prefer this for the following reasons:
- I reviewed the abstracts of lots of RFCs. They are largely one paragraph.
- The abstract should be easy to read, so not so many acronyms
- Some of the things you are trying to in the abstract should be addressed in section 1.
- I very much want to give the concrete example of devices in the abstract because readers won't know what an entity is

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -237,13 +237,11 @@ informative:
 
 An Entity Attestation Token (EAT) provides an attested claims set
 that describes state and characteristics of an entity,
-a device like a phone, IoT device, network equipment or such.  This claims set is used by a
+a device like a smartphone, IoT device, network equipment or such.  This claims set is used by a
 relying party, server or service to determine how much it wishes to trust the entity.
 
 An EAT is either a CBOR Web Token (CWT) or JSON Web Token (JWT) with attestation-oriented
-claims. To a large degree, all this document does is extend
-CWT and JWT.
-
+claims.
 
 --- middle
 


carl-wallace:
I don’t see where 279 answers the mail re: EAT serving as a framework at all, which was most of my goal re: abstract and intro mods. I do agree section 1 should reflect this as well and if you want it all there instead of the abstract, that’s fine. I tried to get the idea into the abstract with an echo in the intro. I agree fewer acronyms is preferrable but still want to get the points across. I will cancel 259 and this can serve as vehicle for abstract mods. If there are other bits from 259 you want to use (like mods for brevity), one of us can submit another PR later.

 

Date: Wednesday, August 31, 2022 at 2:15 PM
Cc: Subscribed ***@***.***>
Subject: [ietf-rats-wg/eat] Finalize abstract (PR #279)

 

I'd like to use this thread to finalize the abstract. I won't merge it until us authors are agreed on the final abstract.

This change addresses a comment by Eliot that EAT is more than just extensions to JWT/CWT.

Carl, I looked at your proposed changes #259 and prefer this for the following reasons:
I reviewed the abstracts of lots of RFCs. They are largely one paragraph.
The abstract should be easy to read, so not so many acronyms
Some of the things you are trying to in the abstract should be addressed in section 1.
I very much want to give the concrete example of devices in the abstract because readers won't know what an entity is
You can view, comment on, or merge this pull request online at:
  https://github.com/ietf-rats-wg/eat/pull/279
Commit Summary
d3ce338 Updates to abstract
File Changes 
(1 file)
M draft-ietf-rats-eat.md (6) 
Patch Links:
https://github.com/ietf-rats-wg/eat/pull/279.patch
https://github.com/ietf-rats-wg/eat/pull/279.diff

laurencelundblade:
Right, this abstract doesn't introduce EAT as a framework at all. Seems not necessary in the abstract and better to do in section 1.  Maybe if we came up with one or two simple sentences it could be in the abstract.

The comment Eliot that this does address is the removal of "To a large degree, all this document does is extend CWT and JWT."

PR #283, opened 2022-09-10
minor edits to the entity overview section

carl-wallace:


draft-ietf-rats-eat.md
```diff
@@ -290,22 +290,13 @@ In particular, it can be used for evidence and attestation results.
 
 ## Entity Overview
 
-The document uses the term "entity" to refer to the target of the attestation token.
-The claims defined in this document are claims about an entity.
-
-An entity is an implementation in hardware, software or both.
```
laurencelundblade, line 296: Could you add this, or equivalent back in?  I think it is important to mention that an entity can be hardware. That doesn't occur to very many people because not many people have ever worked on chip HW.

carl-wallace: Added back the last sentence, the other stuff in the snip above was already there.

PR #286, opened 2022-09-16
Shorter Text about Claims that contain Keys

hannestschofenig:


mcr (approved):

hannestschofenig:
I would either do two things:

1) Remove the section altogether, or 
2) State that this document does not define the functionality since it is not uncommon to also speak about what is left for future work.

laurencelundblade:
OK. Then #311 removes the section entirely. 

(Document is getting shorter :-)

laurencelundblade:
Removed the section entirely

PR #287, opened 2022-09-16
Secure Boot

hannestschofenig:


laurencelundblade:
I think telling a relying party that the SW is under control of a particular organization is much more valuable than that the SW is under control of some unknown organization.  

A lot of this is because we operate in a world where the end owner/user of the device is not trusted to act in their own self interest because they don't know what to do. People using electricity from a nuclear power plant are not trusted to operate the power plant. FIDO authenticators under control of the end user aren't trusted by FIDO RPs.

If you just say secure boot is enabled, the RP probably doesn't know what that means in trusting the device.

hannestschofenig:
The secure-boot claim allows to convey "true" or "false". 

The description, however, does not match what secure boot means. 

You are saying "Secure boot is considered enabled when the firmware and operating system, are under control of the 
manufacturer of the entity identified in the "oemid" claim described in {{oemid}}."

Secure boot, unlike with measured boot, uses unchangable code (typically code in ROM) to verify the signature of a piece of firmware prior to starting that firmware. That new firmware then verifies the next layer of software, etc.

The question for me is therefore what does the "firmware and operating system, are under control of the manufacturer" actually means.

laurencelundblade:
The value this claim provides to the verifier or relying party is knowing the base OS and SW on the device is something they can trust, in particular that it is the SW that the OEM intended the device to run.

It doesn’t matter how the device achieves this for this claim. In most cases it will be as you say, stages of code authentication starting with ROM. It could be that there is only ROM. It could also be that there is no SW. The device is just pure HW.

Perhaps the claim should be renamed to “OEM approved boot”? 

Here’s some use cases where the answer should be “false”.  The boot ROM does SW verification of the next stage and so on up to the OS, but the end user has access to the signing keys and can run whatever they want.  This use of secure boot is a fail if the device does DRM / content protection because the user can circumvent it. Another fail is loading SW into a mobile phone that allows it to manipulate the cellular signals, particular the radio power output.

bremoran:
Which OEM? Is it the silicon vendor? The software vendor? The equipment manufacturer? The white-label vendor? I don't understand what OEM means. Supply chains are complex. There is no one OEM in complex devices. It may be possible for a simple device to contain only a silicon vendor, an equipment manufacturer, and a software vendor, but ANY of these three could be "in control" of the software. This depends on business models more than technical models.

laurencelundblade:
Hi Brendon,

It is the OEM identified by the OEM ID claim.

Whether that vendor is for silicon, the circuit board, the OEM device or the end branded device depends on the scope of the EAT itself.

Agreed that this claim will not work in all circumstances (like most claims) and depends on knowing what the EAT applies to (like most claims).

laurencelundblade:
Selected #340 instead.

PR #298, opened 2022-09-19
Removal of the Transport Security section

hannestschofenig:


gmandyam:
I don't see an issue with merging.  Going ahead with it now.

PR #164, opened 2022-02-07
Add section on privacy considerations for replay protection

gmandyam:

PR #299, opened 2022-09-19
Re-written Privacy Considerations

hannestschofenig:
I changed the privacy consideration section.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1678,47 +1678,48 @@ Nested-Token is defined in the following sections.
 # Privacy Considerations {#privacyconsiderations}
 
 EAT is a building block that can be used in a variety of use cases. The details of the 
-use cases matter for the purpose of evaluating a possible privacy impact. While EAT
+use cases matter for evaluating the privacy impact. While EAT
 does not expose information about individuals but rather about device hardware, and low level
 software (like bootloaders and firmware), the information may be linked to individuals whenever
-there is an individual using the device. Of course, there are many use cases where there is 
-no link to an individual, for example in industrial use cases. 
+there is an individual using the device. There are, however, many use cases where there is 
+no link to an individual, for example in industrial use cases.
 
-The claims in Entity Attestation Tokens can reveal a lot of information about a device.
-Privacy threats fall into three categories, namely 
+The claims in Entity Attestation Tokens can reveal a lot of information about a device, the 
+software running on the device and other information (manufacturer, location). To evaluate
+the privacy implications at least three categories need to be investigated, namely 
 - threats that concern the transports of EATs from the attester to the relying party,
-- secondary use, and 
+- secondary use, 
 - identification and correlation. 
 
 Communication security is used to protect an EAT while in transit so that third parties cannot inspect the content.
 This can be accomplished using classical communication security techniques, such as TLS, but also by using object security, 
 such as COSE encryption. While the use of encryption at the object level is possible via techniques
-offered by COSE, no mechanisms are mandated in this document. Profiles of this specification have to offer
-additional guidance when such object security protection is needed. The use of TLS is, however, a frequently
-used technique and offers comparable properties.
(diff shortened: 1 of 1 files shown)


gmandyam:
This proposed PR removes a section on replay protection and privacy that was reviewed and agreed upon by the Architecture team and EATS editors:  see https://github.com/ietf-rats-wg/eat/pull/164.

hannestschofenig:
The text about replay protection has little to no implementation for privacy and is misplaced in the document.

gmandyam:
> The text about replay protection has little to no implementation for privacy and is misplaced in the document.

Would recommend you take your concerns back to the RATS arch. team who requested this text and reviewed it before it was merged.

Also I did not find this issue raised in the iotdir LC review (https://mailarchive.ietf.org/arch/msg/rats/zubR3clEkOwh68bPvwbD8NPwY2k/) nor secdir LC review (https://mailarchive.ietf.org/arch/msg/secdir/pHag-WP-Zkqo-Yvzv_RdXDfSGsM/).

hannestschofenig:
Good idea, Giri.

hannestschofenig:
Here is what I posted to the RATS list:

> In PR https://github.com/ietf-rats-wg/eat/pull/299 I proposed a re-write of the privacy consideration section.
> 
> As part of my re-write I removed text that was, according to Giri, “reviewed and agreed upon by the Architecture team and EATS editors in https://github.com/ietf-rats-wg/eat/pull/164”. 
> 
> Now, I would like to bring it to the group. Here is the relevant text: 
> 
> 8.4.  Replay Protection and Privacy
> 
>    EAT offers 2 primary mechanisms for token replay protection (also
>    sometimes known as token "freshness"): the cti/jti claim and the
>    nonce claim.  The cti/jti claim in a CWT/JWT is a field that may be
>    optionally included in the EAT and is in general derived on the same
>    device in which the entity is instantiated.  The nonce claim is based
>    on a value that is usually derived remotely (outside of the entity).
>    These claims can be used to extract and convey personally-identifying
>    information either inadvertently or by intention.  For instance, an
>    implementor may choose a cti that is equivalent to a username
>    associated with the device (e.g., account login).  If the token is
>    inspected by a 3rd-party then this information could be used to
>    identify the source of the token or an account associated with the
>    token (e.g., if the account name is used to derive the nonce).  In
>    order to avoid the conveyance of privacy-related information in
>    either the cti/jti or nonce claims, these fields should be derived
>    using a salt that originates from a true and reliable random number
>    generator or any other source of randomness that would still meet the
>    target system requirements for replay protection.
> 
> The RATS architecture talks about three approaches for providing freshness, namely
> - timestamps,
> - nonces, and
> - epoch IDs.
> 
> The text above talks about two mechanisms, namely
> - nonces, and
> - the cti/jti.
> 
> (As you will see later, the cti/jti does not correspond to one of the freshness mechanisms from the RATS architecture.)
> 
> In the EAT draft version -14 the cti/jti claims are mentioned in two sections, namely in Section 8.4 (see above) and also in Section 4.3.1. The cti claim contains a unique identifier for the JWT. 
> 
> Assuming that an implementer uses the cti to convey a username / account login is unjustified given what Section 4.1.7 of RFC 7519 defines it to be, see  https://www.rfc-editor.org/rfc/rfc7519#section-4.1.7.
> 
> So, there is only a privacy problem with the cti/jti if you use it in a way that has not been envisioned and even suggested by the RFC that defined it.
> 
> The privacy implications of the nonce are not described nor are other privacy implications of the iat (issued at) claim defined, which would correspond to the timestamp replay protection mechanism defined in the RATS architecture. 
> 
> For this reason I suggested to remove this text from the privacy consideration section. 
>

gmandyam:
This PR is not really applicable given the changes that came in after -14.  If there are suggested changes to the nonce considerations in either the Privacy or Security cons sections then please propose them to the WG to obtain consensus.

henkbirkholz:
@hannestschofenig is there anything left unaddressed here?

PR #304, opened 2022-09-24
Mods to section 3 Top-Level Token Definition

carl-wallace:
This should address all comments relative to similar earlier text except for two conflicting comments related to UCCS reference (Thomas wanted the reference in, Lawrenece wanted the reference out). This version omits it (but as a non-normative reference I don't think including it would slow anything down).

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -417,9 +417,9 @@ Endorsement:
 
 # Top-Level Token Definition
 
-An EAT MUST contain a claims set about an entity packaged as a CWT {{RFC8392}} or JWT {{RFC7519}}. This specification extends {{RFC8392}} and {{RFC7519}} by defining additional claims for attestation. All definitions, requirements, creation and validation procedures, security considerations, IANA registrations, etc. from CWT and JWT apply to the EAT format.
+An EAT is a "message", a "token", or such whose content is a Claims-Set about an entity or some number of entities. An EAT MUST always contains a Claims-Set.
 
-Authenticity and integrity protection MUST be provided for EATs. This document defines the use of CWT or JWT for this purpose.
+Authenticity and integrity protection MUST be provided for EATs. This document relies on CWT or JWT for this purpose.
 Extensions to this specification MAY use other methods of protection.
 
 The identification of a protocol element as an EAT follows the general conventions used for CWTs and JWTs.
@@ -432,8 +432,8 @@ This document also defines also defines a new top-level message, the detached EA
 Detached EAT bundles can be either CBOR or JSON encoded.
 
 The following CDDL defines the top-level `$$EAT-CBOR-Tagged-Token`, `$$EAT-CBOR-Untagged-Token` and `$$EAT-JSON-Token-Formats` sockets, enabling future token formats to be defined.
-Any new format that plugs into one or more of these sockets MUST be defined in a IETF standards track document.
-Of particular use may be a token type that provides no direct authenticity or integrity protection for use with transports mechanisms that do provide the necessary security services {{UCCS}}.
+Any new format that plugs into one or more of these sockets MUST be defined by an IETF standards action.
+Of particular use may be a token type that provides no direct authenticity or integrity protection for use with transports mechanisms that do provide the necessary security services.
 
 Nesting of EATs is allowed and defined in {{Nested-Token}}.
 This includes the nesting of an EAT that is a different format than the enclosing EAT.
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -206,6 +206,7 @@ informative:
 
   CBOR.Cert.Draft: I-D.ietf-cose-cbor-encoded-cert
 
+  UCCS: I-D.ietf-rats-uccs
 
 --- abstract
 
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-eat.md
```diff
@@ -417,23 +417,26 @@ Endorsement:
 
 # Top-Level Token Definition
 
-An EAT is a "message", a "token", or such whose content is a Claims-Set about an entity or some number of entities.
-An EAT MUST always contains a Claims-Set.
+An EAT MUST contain a claims set about an entity packaged as a CWT {{RFC8392}} or JWT {{RFC7519}}. This specification extends {{RFC8392}} and {{RFC7519}} by defining additional claims for attestation. All definitions, requirements, creation and validation procedures, security considerations, IANA registrations, etc. from CWT and JWT apply to the EAT format.
 
-An EAT may be encoded in CBOR or JSON as defined here.
-While not encouraged, other documents may define EAT encoding in other formats.
+Authenticity and integrity protection MUST be provided for EATs. This document defines the use of CWT or JWT for this purpose.
```
laurencelundblade, line 422: Prefer "relies on" rather than "defines the use of".

draft-ietf-rats-eat.md
```diff
@@ -417,23 +417,26 @@ Endorsement:
...
 
-This document also defines the detatched EAT bundle ({{DEB}}), a bundle of some detached Claims-Sets and CWTs or JWTs that provide protection for the detached Claims-Set.
+This document also defines also defines a new top-level message, the detached EAT bundle (see {{DEB}}), which holds a collection of detached claims sets and an EAT that provides integrity and authenticity protection for them.
+Detached EAT bundles can be either CBOR or JSON encoded.
 
-The following CDDL defines the top-levels of an EAT token as a socket indicating future token formats may be defined.
-Any new format that plugs into this socket MUST be defined in a IETF standards track document.
-See {{CDDL_for_CWT}} for the CDDL definitions of a CWT and JWT.
+The following CDDL defines the top-level `$$EAT-CBOR-Tagged-Token`, `$$EAT-CBOR-Untagged-Token` and `$$EAT-JSON-Token-Formats` sockets, enabling future token formats to be defined.
+Any new format that plugs into one or more of these sockets MUST be defined in a IETF standards track document.
```
laurencelundblade, line 435: "by an IETF standards action" rather than "in an IETF standards track document"?

draft-ietf-rats-eat.md
```diff
@@ -417,23 +417,26 @@ Endorsement:
...
+Of particular use may be a token type that provides no direct authenticity or integrity protection for use with transports mechanisms that do provide the necessary security services {{UCCS}}.
```
laurencelundblade, line 436: I'm fine with the reference as long as it doesn't slow down EAT.

We had this big discussion in Vienna that I thought concluded that the reference had to go because I didn't think you could refer to a draft document in a published RFC, normative or non-normative. If that's true then with this reference EAT would have to wait for UCCS to become an RFC. 

(And if we can reference UCCS, then maybe we should put back all the CDDL and other stuff for UCCS. There was a fair bit of it. The PR to remove the UCCS reference was large and complicated).

So first question to answer is whether a publish RFC can references an I-D.

draft-ietf-rats-eat.md
```diff
@@ -417,23 +417,26 @@ Endorsement:
 
 # Top-Level Token Definition
 
-An EAT is a "message", a "token", or such whose content is a Claims-Set about an entity or some number of entities.
-An EAT MUST always contains a Claims-Set.
+An EAT MUST contain a claims set about an entity packaged as a CWT {{RFC8392}} or JWT {{RFC7519}}. This specification extends {{RFC8392}} and {{RFC7519}} by defining additional claims for attestation. All definitions, requirements, creation and validation procedures, security considerations, IANA registrations, etc. from CWT and JWT apply to the EAT format.
```
laurencelundblade, line 420: I would like to retain my original two sentences here. I think they provide a crucial definition of an EAT that goes beyond CWT and JWT.  They define what can plug into the CDDL sockets mention below and what cannot.

I think it is important to have some text that says what an EAT is independent of it being a CWT/JWT.

Most of this paragraph is repetition of the opening paragraph in section 4. I'm not sure whether it is better here or in section 4, but it shouldn't be in both.

I think the text say what an EAT is independent of CWT/JWT should be separate of how EAT relates to a CWT/JWT.

laurencelundblade (changes requested):
Some discussion / resolution needed on how / what we describe an EAT here.

Fine with UCCS ref, but it looks like the target of {{UCCS}} is missing.

draft-ietf-rats-eat.md
carl-wallace: I agree when editing section by section it is difficult to control repetition. Will restore your two sentences though it is not clear to me how in the first two sentences of the change fail to get the point across (though "an entity" should have been "an entity or a set of entities".

draft-ietf-rats-eat.md
carl-wallace: see above

draft-ietf-rats-eat.md
carl-wallace: removed as was intent, I must have missed a save

draft-ietf-rats-eat.md
hannestschofenig: If you consider UCCS, then these two sentences are not correct since with UCCS integrity and authenticity is not provided by the CWT/JWT

carl-wallace: The next sentence is "Extensions to this specification MAY use other methods of protection."

draft-ietf-rats-eat.md
```diff
@@ -417,23 +417,26 @@ Endorsement:
...
-EAT as defined here is always integrity and authenticity protected through use of CWT or JWT.
-Other token formats using other methods of protection may be defined outside this document.
+The identification of a protocol element as an EAT follows the general conventions used for CWTs and JWTs.
+Identification depends on the protocol carrying the EAT.
+In some cases it may be by media type (e.g., in a HTTP Content-Type field).
+In other cases it may be through use of CBOR tags.
+There is no fixed mechanism across all use cases.
 
-This document also defines the detatched EAT bundle ({{DEB}}), a bundle of some detached Claims-Sets and CWTs or JWTs that provide protection for the detached Claims-Set.
+This document also defines also defines a new top-level message, the detached EAT bundle (see {{DEB}}), which holds a collection of detached claims sets and an EAT that provides integrity and authenticity protection for them.
```
laurencelundblade, line 431: "also defines" is duplicated

draft-ietf-rats-eat.md
laurencelundblade: Actually, I'd failed to save a major change to my comment above that said it was OK to leave it in. ooops.

I found that you can have a non-normative reference to a I-D. The method for doing it is here: https://www.rfc-editor.org/rfc/rfc7322#section-4.8.6.4

I am OK either way. It might be kind of good to make the reference so the reader knows about it rather than hoping they'll come across it.

draft-ietf-rats-eat.md
laurencelundblade: The first sentence of your changes says to me that and EAT MUST be a CWT or JWT implying that nothing else can be an EAT.

draft-ietf-rats-eat.md
laurencelundblade: Seems OK to me now.

draft-ietf-rats-eat.md
carl-wallace: Will put it back.

draft-ietf-rats-eat.md
carl-wallace: Fixed.

draft-ietf-rats-eat.md
carl-wallace: I already put your two sentences back, so all good.

laurencelundblade (approved):

Issue #294, opened 2022-09-17
what is purpose of tolerating jti/cti claims


mcr:
> 13) 4.3.1. Token ID Claim (cti and jti) Should it even be present?

A sentence was removed, but it still seems to me that cti and jti MUST be
ignored if present, and SHOULD not be sent.

gmandyam:
There are several instances where an attestation is required but a nonce is not possible due to transport limitations (e.g. one-way communication such as BLE-AD).  A freshness mechanism supported by the token issuer is better than nothing in these situations.

If the intent with this issue  is to cover the case where both a cti and nonce are sent (CWT case), then the cti cannot be ignored if it included in the payload used for generation of the signature.  It could be ignored for the purpose of freshness assessment in that case, but the text states "MUST be ignored if present" which is not the same.

laurencelundblade:
At this point we exclude no claims from EAT for any reason. Anything in the IANA registries is fair game. Seems, we'd have to have a really good reason to explicitly, universally and categorically exclude a particular claim.

Do you have a particular problem in mind? I can't think of anything myself.

I don't think Giri's freshness example is the only use for CTI/JTI. Someone might just want a good unique ID for tracking tokens in a database that doesn't depend on anything else like time, device ID, ...

gmandyam:
>I don't think Giri's freshness example is the only use for CTI/JTI. Someone might just want a good unique ID for tracking tokens in a database that doesn't depend on anything else like time, device ID, ...

CTA WAVE's Common Access Token, which has been presented at the IETF CDNI WG (https://datatracker.ietf.org/meeting/114/materials/agenda-114-cdni-01), makes use of cti,  In this case, the CWT issuer is distinct from the client that actually relays the token and is never in direct contact with the relying party (in this case a Content Delivery Network).  

There may be comparable cases such as delegated attestation, where the attester is not integrated with the entity but the entity must relay the token on behalf of the attester to the relying party.

mcr:
Okay, let me ask the question differently: why are jti/cti mentioned at all?

carl-wallace:
#308 removes this section, since cti/jti can be used without this section and there are not limitations on use a la those for iat.

laurencelundblade:
Claims where removed

PR #308, opened 2022-09-24
remove cti/jti section. 

carl-wallace:
Also modify a few cti/jti references for consistency. The removal of the cti/jti section addresses issue #294.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1240,13 +1240,6 @@ They may appear in evidence or attestation results.
 When these claims appear in evidence, they SHOULD not be passed through the verifier into attestation results.
 
 
-### cti and jti (Token ID) Claims
-
-CWT defines the "cti" claim. JWT defines the "jti" claim. These are
-equivalent in EAT and carry a unique token identifier as
-they do in JWT and CWT.
-
-
 ### iat (Timestamp) Claim
 
 The "iat" claim defined in CWT and JWT is used to indicate the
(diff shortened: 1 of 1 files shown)


gmandyam:
https://datatracker.ietf.org/doc/html/draft-ietf-rats-eat#section-8.4 was added in at the request of the RATS arch. team.  If EAT is to remove all mention of cti/jti then this should likely be revised.  Would suggest the following modified Privacy Section text to be added to the PR:

" EAT defines the nonce claim for token replay protection (also sometimes known as token "freshness"). The nonce claim is based on a value that is usually derived remotely (outside of the entity).  This claim can be used to extract and convey personally-identifying information either inadvertently or by intention.  For instance, an implementor may choose a nonce that is equivalent to a username associated with the device (e.g., account login).  If the token is  inspected by a 3rd-party then this information could be used to identify the source of the token or an account associated with the token.  In order to avoid the conveyance of privacy-related information in the nonce claim, it should be derived using a salt that originates from a true and reliable random number generator or any other source of randomness that would still meet the target system requirements for replay protection."

carl-wallace:
Changed to use the suggested text.

 

From: Giri Mandyam ***@***.***>
Date: Sunday, September 25, 2022 at 2:18 PM
Cc: Carl Wallace ***@***.***>, Author ***@***.***>
Subject: Re: [ietf-rats-wg/eat] remove cti/jti section. (PR #308)

 

https://datatracker.ietf.org/doc/html/draft-ietf-rats-eat#section-8.4 was added in at the request of the RATS arch. team. If EAT is to remove all mention of cti/jti then this should likely be revised. Would suggest the following modified Privacy Section text to be added to the PR:

" EAT defines the nonce claim for token replay protection (also sometimes known as token "freshness"). The nonce claim is based on a value that is usually derived remotely (outside of the entity). This claim can be used to extract and convey personally-identifying information either inadvertently or by intention. For instance, an implementor may choose a nonce that is equivalent to a username associated with the device (e.g., account login). If the token is inspected by a 3rd-party then this information could be used to identify the source of the token or an account associated with the token. In order to avoid the conveyance of privacy-related information in the nonce claim, it should be derived using a salt that originates from a true and reliable random number generator or any other source of randomness that would still meet the target system requirements for replay protection."

PR #309, opened 2022-09-26
Add reference to Claims-Set-Claims socket to intro of section 4.

carl-wallace:
Otherwise Claims-Set-Claims only appears in CDDL.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -448,6 +448,8 @@ CDDL, along with a text description, is used to define each claim
 independent of encoding.  Each claim is defined as a CDDL group.
 In {{encoding}} on encoding, the CDDL groups turn into CBOR map entries and JSON name/value pairs.
 
+Each claim is appended to the `$$Claims-Set-Claims` socket. Specifications that define additional claims SHOULD also supply additions to the `$$Claims-Set-Claims` socket.
+
 All claims in an EAT MUST use the same encoding except where otherwise explicitly stated (e.g., in a CBOR-encoded token, all claims must be CBOR-encoded).
 
 This specification includes a CDDL definition of most of what is defined in {{RFC8392}}.


draft-ietf-rats-eat.md
```diff
@@ -448,6 +448,8 @@ CDDL, along with a text description, is used to define each claim
 independent of encoding.  Each claim is defined as a CDDL group.
 In {{encoding}} on encoding, the CDDL groups turn into CBOR map entries and JSON name/value pairs.
 
+Each claim is appended to the `$$Claims-Set-Claims` socket. Specifications that define additional claims SHOULD also supply additions to the `$$Claims-Set-Claims` socket.
```
hannestschofenig, line 451: Under what circumstances should those specifications not supply such additions?

carl-wallace: Perhaps it should be a MUST. I was thinking of claims that only had some narrow or local significance, but even in that context the claim would be added to the socket. Will change it.

laurencelundblade: I think someone defining future EAT claims can use CDDL or not use CDDL. I don't think we want to mandate CDDL. It is OK to use only prose to define an additional claim for people that don't like or don't know CDDL. That's when $$Claims-Set-Claims is not used.

This addition and the use of the word "SHOULD" here seems right to me.

I think I'd say something like "all claims are defined in this document in CDDL and added to the $$Claims-Set-Claims socket group." Don't think "append" is the right word. Then somethign like "claims added outside this document SHOULD also add to the $$Claims-Set-Claims socket group, but may also defined claims in prose."

carl-wallace: The previous sentence states that "Each claim is defined as a CDDL group", so I don't think we need to repeat that. As to whether or not someone writes prose or CDDL, a CDDL definition *could* be derived from the prose.  Whether or not someone writes down an addition to Claims-Set-Claims, logically that is what happens anyway (so it's a MUST practically by definition). I don't think making an allowance for prose only here adds anything.

The main goal of this PR is to reference Claims-Set-Claims in the prose of the spec (since it's all over the CDDL). Will use some of the words above to change from append to add and see if that works.

laurencelundblade: I'd like to hear other comments on this one.

This document exclusively uses CDDL and $$Claims-Set-Claims, but CWT and JWT don't and are fine. CDDL is just one way to specify something. Just because the EAT document uses it doesn't mean that follow on documents do. CWT didn't say all claims must be defined in prose.

The actual structure and code implementing EAT is not really affected by whether CDDL or prose or some other definition language is used.

I can live with requirement that CDDL MUST be used, but much prefer SHOULD.

PR #318, opened 2022-10-03
bundles are one encoding format

laurencelundblade:
This clarifies that all detached claims sets in a bundle must be the same encoding as the bundle.

That is if the detached EAT bundle is CBOR-encoded, then all the detached claims-sets in it must be CBOR-encoded (and bstr wrapped).  The token with the detached digests doesn't have to be the same though. A sort of working principle here is that in EAT encoding can only be different for whole tokens.

We could make an exception for bundles, but we'd need more protocol elements for JSON-encoded tokens so they could distinguish a JSON-detached claims-set from a CBOR-detached claims set. (The same is not needed for CBOR because the tstr/bstr could be used).

There's no requirement that detached Claims-Sets be carried in a detached EAT bundle. They could be in a multipart MIME for example.  In that case mixing is fine because there are media types.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1315,14 +1315,14 @@ It is a top-level EAT message like a CWT or JWT.
 It can be occur any place that CWT or JWT messages occur.
 It may also be sent as a submodule.
 
-A detached EAT bundle has two main parts.
+A detached EAT bundle consists of two parts.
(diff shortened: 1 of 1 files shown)


carl-wallace (commented):
I'd change to MUST be XXX-encoded since the next sentence reiterates the point with a MUST.

PR #338, opened 2022-11-16
Clarify Attestation

### LINKED ISSUE #337: What are "attestations"
Originally raised by Henk during IETF 115 meeting:
Section 4.3.3 has
> 4 -- Certificate Issuance:
Certification Authorities (CA's) may require attestations prior to the issuance of certificates related to keypairs hosted at the entity. An EAT may be used as part of the certificate signing request (CSR)."

(note "attestations" in that text)

I suspect above means "Attestation Results" instead of "attestations"

hannestschofenig:
Fix #337

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1221,7 +1221,7 @@ consumer.  Examples include key material or device management trees.  The consum
 may require an EAT to assess entity security state of the entity prior to provisioning.
 
 4 -- Certificate Issuance:
-: Certification Authorities (CA's) may require Evidence and Attestation Results to make 
+: Certification Authorities (CA's) may require Attestation Results (which in a Background Check model might require receiving Evidence to be passed to a Verifier) to make 
 decisions about the issuance of certificates.  An
 EAT may be used as part of the certificate signing request (CSR).
 
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1221,8 +1221,8 @@ consumer.  Examples include key material or device management trees.  The consum
 may require an EAT to assess entity security state of the entity prior to provisioning.
 
 4 -- Certificate Issuance:
-: Certification Authorities (CA's) may require attestations prior to
-the issuance of certificates related to keypairs hosted at the entity.  An
+: Certification Authorities (CA's) may require Evidence and Attestation Results to make 
+decisions about the issuance of certificates.  An
 EAT may be used as part of the certificate signing request (CSR).
 
 5 -- Proof-of-Possession:


carl-wallace:
Looks good to me except I'd not capitalize the terms Evidence and Attestation Results. There was feedback in ~July that resulted in blanket changes to the document to move away from capitalized terms like these.

draft-ietf-rats-eat.md
```diff
@@ -1221,8 +1221,8 @@ consumer.  Examples include key material or device management trees.  The consum
 may require an EAT to assess entity security state of the entity prior to provisioning.
 
 4 -- Certificate Issuance:
-: Certification Authorities (CA's) may require attestations prior to
-the issuance of certificates related to keypairs hosted at the entity.  An
+: Certification Authorities (CA's) may require Evidence and Attestation Results to make
```
thomas-fossati, line 1224:
```suggestion
: Certification Authorities (CA's) may require evidence and attestation results to make 
```

dthaler: I disagree with the use of Evidence here.   If it is intended to be correct, then something should explain why since this is against WG consensus on the arch doc.  It should just say "Attestation Results" in my view.

carl-wallace: I would expect the CA to receive evidence as a general rule. Re-reading that sentence, it should probably say "evidence or attestation results" but I'd not change it to attestation results alone. See https://datatracker.ietf.org/doc/html/draft-bweeks-acme-device-attest-01 and https://datatracker.ietf.org/doc/html/draft-ietf-lamps-key-attestation-ext-00 for specs that will convey evidence to a CA.

thomas-fossati: For example, in Veracruz, the CA receives evidence (embedded in a CSR, IIRC) extracts it and forwards it to the "background check" verifier.  The same model is used in [ACME's key attestation](https://datatracker.ietf.org/doc/draft-bweeks-acme-device-attest/01/), where the webauthn "attestation object" ends up in the verification challenge reply.

dthaler: Thomas's point is that the CA may "receive" evidence as opaque and forward it on to a verifier in a background check scenario.  It then "receives" as non-opaque, and processes, the attestation results.  I was interpreting receives as the latter sense, whereas I think Carl might be using it in the former sense.   However, the text in the PR does not use the word "receives", it uses the word "requires", which to me is stronger in the former sense.   But it would be best to be explicit if you're referring to the Background Check model, such as

```suggestion
: Certification Authorities (CA's) may require Attestation Results (which in a Background Check model might require receiving Evidence to be passed to a Verifier) to make 
```

laurencelundblade: I don't really think we want to mention attestation evidence and the background check model every time an RP wants/requires/gets/needs attestation results.

laurencelundblade, line 1224: Definitely should be lower case "attestation results" and other terms.

PR #340, opened 2022-11-25
Rename secboot --> oemboot

laurencelundblade:
This is per discussion at IETF 115.

This is a change to the semantics of a claim that was pre-registered. The CBOR integer claim key was not changed, but the JSON claim name was changed from "secboot" to "oemboot". 

This is an alternate to #287


draft-ietf-rats-eat.md
```diff
@@ -732,18 +732,18 @@ A full CoSWID manifest or other type of manifest can be instead if this is too s
...
+### oemboot (OEM Authorized Boot) Claim
 
-A "secboot" claim with value of true indicates secure boot is enabled. Secure boot is
-considered enabled when the firmware and operating
-system, are under control of the manufacturer of the entity identified in the
-"oemid" claim described in {{oemid}}.
-Control by the manufacturer of the firmware and the operating system may be by it being in ROM, being cryptographically authenticated, a combination of the two or similar.
+An "oemboot" claim with value of true indicates the entity booted with software authorized by the manufacturer of the entity as indicated by the "oemid" claim described in {{oemid}}.
+It indicates the firmware and operating system are fully under control of the OEM and may not be replaced by the end user or even the enterprise that owns the device.
+The means of control may be by cryptographic authentication of the software, by the software being in ROM, a combination of the two or other.
```
nedmsmith, line 739: The "or other" leaves the door wide open for any other interpretation which seems odd given the rest of the text is fairly prescriptive. If the CDDL is more constrained than this, then possibly the description should not include "or other"?

carl-wallace: The last sentence lists examples. Would changing the last sentence as follows address your concern: "Example means of control include cryptographic authentication of the software, the software being in ROM, or a combination of the two."?

PR #346, opened 2022-12-09
Fix wording of CA

hannestschofenig:


laurencelundblade:
Hi Hannes,

I switched to Certificate Authority because that is the term defined in RFC 4949 and because Certification Authority might be confused with something like a performer of Common Criterion certification.

But if there's consensus around Certification Authority, I'm OK with that. It's a minor point in the document as a whole.

carl-wallace:
I think certification authority is more commonly used and is what is used in RFC 5280.

 

Date: Friday, December 9, 2022 at 11:32 AM
Cc: Subscribed ***@***.***>
Subject: Re: [ietf-rats-wg/eat] Fix wording of CA (PR #346)

 

Hi Hannes,

I switched to Certificate Authority because that is the term defined in RFC 4949 and because Certification Authority might be confused with something like a performer of Common Criterion certification.

But if there's consensus around Certification Authority, I'm OK with that. It's a minor point in the document as a whole.

henkbirkholz:
RFC4949 says:

> $ certificate authority
>       (D) Synonym for "certification authority".
>
>       Deprecated Term: IDOCs SHOULD NOT use this term; it suggests
>       careless use of the term "certification authority", which is
>       preferred in PKI standards (e.g., [[X509](https://www.rfc-editor.org/rfc/rfc4949#ref-X509), [R3280](https://www.rfc-editor.org/rfc/rfc4949#ref-R3280)]).

gmandyam:
"Certification Authority (-ies)" is used in PKIX docs such as https://datatracker.ietf.org/doc/rfc2459/.  It makes sense to remain consistent with IETF terminology.  Would recommend using "Certification Authority (CA)" in first use of the term, and only using CA for the rest of the document.

henkbirkholz:
+1 Giri, obviously ☺️

Issue #343, opened 2022-12-07
inconsistency in `iat` definition


thomas-fossati:
The CDDL allows float (in fact, `time` is `int / float`) but [§4.3.1](https://ietf-rats-wg.github.io/eat/draft-ietf-rats-eat.html#section-4.3.1) disallows it.

We should use `int` rather than `~time` in the CDDL definition.

carl-wallace:
The definition traces to Section 2 of RFC8392 then to Section 2.4.1 of RFC 7049. Given the data type for this reused claim is established elsewhere, profiling its use in prose seems correct to me (vs using an alternative definition). An implementation that does not reuse a data type from a CWT library may elect to enforce the constraint as you suggest.

laurencelundblade:
Since most problems in CS can be solved with a level of indirection :-) we should use ~time-int, not int.

I don't want to change the CDDL in Appendix D because it is copied from draft UCCS and represents a CWT and JWT. So don't want to change the iat claim there.  But because you might miss the text in section 4.3.1, I added a paragraph to appendix D in this PR pointing it out.

I'm also OK leaving it as is as Carl suggests.

The problem with both these solutions is that CDDL with a floating-point iat will validate. The only solution I know of to that is to change the CDDL for the CWT. I suppose we could do that and put a note in that we changed it.

thomas-fossati:
If CDDL can express a constraint and we don't use it we dilute its usefulness.

I'm not going to insist, do what you think it's the right thing to do.

carl-wallace:
(author hat off/personal opinion hat on)

 

Ultimately, CDDL will have import as well, as I understand it. That’s why I suggested we accept the definition as-is and profile its use in prose (since importing from CWT is what is happening). With different prose and definition, it really isn’t reuse of iat. 

 

From: cabo ***@***.***>
Reply-To: ietf-rats-wg/eat ***@***.***>
Date: Monday, December 12, 2022 at 7:05 AM
To: ietf-rats-wg/eat ***@***.***>
Cc: Carl Wallace ***@***.***>, Comment ***@***.***>
Subject: Re: [ietf-rats-wg/eat] inconsistency in `iat` definition (Issue #343)

 

CDDL has .and (and .within) to make further constraints.
The authors should consider whether this is appropriate here.

hannestschofenig:
I would suggest to use as simple CDDL constructs as possible right now. Later, when CDDL gets enhanced more you can make this better.

[...]

PR #358, opened 2022-12-10
Add sentence about ~time-int for iat claim in Appendix D

laurencelundblade:
Proposed fix for #343.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -154,7 +154,7 @@ normative:
     target: <https://www.iana.org/assignments/core-parameters>
 
 
-  SUIT.Manifest: I-D.draft-ietf-suit-manifest
+  SUIT.Manifest: I-D.ietf-suit-manifest
 
 
 informative:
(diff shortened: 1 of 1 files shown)


dthaler (approved):

draft-ietf-rats-eat.md
```diff
@@ -2364,6 +2364,8 @@ EAT's definition of permanence is in terms of operations and device lifecycle.
 Following is CDDL specifying CWT as it is needed to complete this specification.
 This CDDL also covers the Claims-Set for JWT.
 
+Note that {{iat-claim}} requires that the iat claim be the type ~time-int, not the type ~time when it is used in an EAT as floating-point values are not allowed for the "iat" claim in EAT.
```
thomas-fossati, line 2367:
```suggestion
Note that {{iat-claim}} requires that the iat claim be the type ~time-int ({{common-types}}), not the type ~time when it is used in an EAT as floating-point values are not allowed for the "iat" claim in EAT.
```

thomas-fossati (approved):
LGTM

(I left a small editorial suggestion inline)

PR #360, opened 2022-12-13
Remove sentence about section 4 claims passing through

laurencelundblade:
The idea is to be silent on this because who ever builds a system that might or might not pass these claims through will probably know more than we do.

To me, there are clearly cases were these claims might pass through and clearly cases where they won't. For example the verifier might be a light-touch verifier only doing the signature verification so it makes sense to pass through intended use and profile. In other cases the evidence token and attestation results token might be completely different in nature. I don't see what value we can add by speculating about this.

The discussion in 1.3, particularly 1.3.1 applies here. I think that's all we need to say.

I don't see any security risks by being silent here. There's no obvious attacks that might happen if these are passed through.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -2518,10 +2518,8 @@ The following is a list of known changes since the immediately previous drafts. 
 non-authoritative.  It is meant to help reviewers see the significant
 differences. A comprehensive history is available via the IETF Datatracker's record for this document.
 
-## From draft-ietf-rats-eat-17
-- Rename secboot to oemboot and describe it as OEM Authorized Boot
-- Replace use of "attestations" in intended use claim
-- Be clear about nonce providing freshness and replay protection in privacy considerations section
+## From draft-ietf-rats-eat-18
+- Remove sentence discussing pass through of claims about the token in section 4.3
 
 --- contributor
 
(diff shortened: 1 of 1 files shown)


thomas-fossati:
> I don't see any security risks by being silent here. There's no obvious attacks that might happen if these are passed through.

Not fully sure about this.  E.g., If `iat` is used as an input to the appraisal policy and there is confusion about what it refers to (evidence generation vs evidence appraisal), then - in principle - there may be a "security" problem.

laurencelundblade:
Yeah, a mixed up iat could be a security issues, but to me anyone implementing EAT, particularly anyone creating a profile and/or creating a key distribution/identification system for EAT will face a many issues like this that we can't do anything about. We've kind of already assumed they are this smart. Also don't think that token ID's are a subtle issue that they might miss compared to others like those about the use of crypto that we often put in security considerations.

thomas-fossati (approved):

thomas-fossati:
I'd be happier with better / safer recommendations (e.g., "if you need to passthrough, then clearly separate contexts -- e.g., by isolating them in their own sub-map").  However, I can live with just being silent and hope that the reader is a competent person.

laurencelundblade:
Maybe someday we do an information RFC that is more of a cook book guidance on some of the ways that EATs get nested, ID'd... There's clearly a lot to it. Appreciate your comments, Thomas!

PR #364, opened 2023-02-06
address base64url padding issue raised by Thomas

carl-wallace:
See https://mailarchive.ietf.org/arch/msg/rats/qR-udRdc0kFx3ZeGeiypmYv8su0/

### DIFF
--- a/cddl/common-types.cddl
+++ b/cddl/common-types.cddl
@@ -6,7 +6,7 @@ time-int = #6.1(int)
 ; binary data that works for both JSON and CBOR. 
 binary-data = JC< base64-url-text, bstr>
 
-base64-url-text = tstr .regexp "[A-Za-z0-9_=-]+"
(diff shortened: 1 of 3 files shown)


henkbirkholz (approved):

laurencelundblade:
+1

carl-wallace:
\\ is what is used in Section 3.8.3 of RFC8610. Doesn't matter much to me which way it goes though. This came up while trying the updated cddlc (even though this may not be new in that tool). [.] also makes cddlc happy too.

thomas-fossati:
> \ is what is used in Section 3.8.3 of RFC8610. Doesn't matter much to me which way it goes though. This came up while trying the updated cddlc (even though this may not be new in that tool). [.] also makes cddlc happy too.

I'll leave it at your (editor) discretion.  I am happy either way.

PR #372, opened 2023-04-07
Fixes for manifest section and manifest type registrations

laurencelundblade:


draft-ietf-rats-eat.md
```diff
@@ -895,14 +895,14 @@ Installation is anything that adds software to the entity, possibly factory inst
 The defining characteristic is they are created by the software manufacturer.
 The purpose of these claims in an EAT is to relay them without modification to the verifier and possibly to the relying party.
 
-Some manifests may be signed by their software manufacturer before they are put into this EAT claim.
-When such manifests are put into this claim, the manufacturer's signature SHOULD be included.
-For example, the manifest might be a CoSWID signed by the software manufacturer, in which case the full signed CoSWID should be put in this claim.
+Some manifests are signed by their software manufacturer independently, and some are not either because they do not support signing or the manufacturer chose not to sign them.
+For example, a CoSWID might be signed independently before it is included in an EAT.
+When signed manifests are put into an EAT, the manufacturer's signature SHOULD be included even though an EAT's signature will also cover the manifest.
```
carl-wallace, line 900: This doesn't really address the comment from the AD review "If the manufacturer's signature isn't included, how would one know that it is signed?" Maybe delete this sentence and restate to say a signature MAY be removed.

Manifest signatures MAY be removed in cases where an EAT's signature is deemed sufficient.

laurencelundblade: I don't think it is necessary for the consumer to know that the manifest had a signature and it was removed when it was put I the EAT. We don't need any protocol items or even any discussion for that. All this is trying to say is is that the manifest signature is valuable even though EAT is signed and that sometimes the only thing will be the EAT signature.

carl-wallace: OK. Then drop the mention of signatures and leave that to CoSWID (or other). That also would address the comment. It seems odd to me to address the comment "If the manufacturer's signature isn't included, how would one know that it is signed?" with a sentence that beings "When signed manifests are put into an EAT" and concludes by noting the signature might be stripped. If the signature is not there, how does one know "a signed manifest" was put into the EAT?

PR #374, opened 2023-04-09
Remove suggestion about debug claims absence

laurencelundblade:
This is to address Roman's AD review comment.

The alternative would be to add text like this:

> Knowing debug status may be very important for some high-security use cases. If the dbgstat
> claim is absent, those use cases may consider assuming absence means that the debug facilities are enabled.


carl-wallace:
The proposed deletion seems sufficient to me.

PR #381, opened 2023-04-10
Add further info on DLOA retrieval security

laurencelundblade:
This is to address AD review comment.

(I had to review the DLOAs document myself to be sure all was well. It was. This is only describes how the security is achieved. No change).

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -882,6 +882,9 @@ The second element MUST be a platform label indicating which platform was certif
 If the DLOA applies to an application, then the third element is added which MUST be an application label.
 The method of constructing the registrar URI, platform label and possibly application label is specified in {{DLOA}}.
 
+The retriever of a DLOA SHOULD follow the recommendation in {{DLOA}} and use TLS to be sure the DLOA registrar they are accessing is authentic.
+The platform and application labels in the claim indicate the correct DLOA for the entity.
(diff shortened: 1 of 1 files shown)


carl-wallace (commented):
The TLS recommendation in the DLOA spec specifically references mutually authenticated TLS for use in some cases. Ought that be a SHOULD NOT in the EAT context?

laurencelundblade:
The DLOA download is made by the Verifier or RP. Can't think of why that would be SHOULD NOT for them.

carl-wallace:
OK. Then probably worth mentioning that the consumer of the claim MAY be able to mutually authenticate to the resource to obtain the DLOA.

laurencelundblade:
TLS server authentication here is so that the Verifier knows the DLOA is authentic. That's really important so we're just mentioning it so thoughtful readers will have that question answered without going to the DLOA document. The actual requirement comes from DLOA.

I don't think we need to mention client auth here one way or the other. It's not critical to the security of the DLOA and the DLOA document covers it. (The TLS client authentication here is in case the DLOA registrar wishes to restrict who can see DLOAs or a particular DLOA. Since many products that might have DLOAs proudly publish there certifications in the product literature, this doesn't seem important to mention. It is discussed in the DLOA document)

I think the text is fine as is.

carl-wallace:
OK

PR #385, opened 2023-04-13
Improved bundle definition and improved use of the terms 'token' and 'message'

laurencelundblade:
"Message" is used very sparingly to help describe what a token and for a detached EAT bundle.

"EAT" and "token" are used interchangeably and refer to a CWT or JWT (and a UCCS and UJCS).

Removed most use of "full token".

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -199,7 +199,7 @@ claims.
 
 # Introduction
 
-An Entity Attestation Token (EAT) is a message or token made up of claims about an entity.
+An Entity Attestation Token (EAT) is a message made up of claims about an entity.
 An entity may be a device, some hardware or some software.
 The claims are ultimately used by a relying party who decides if and how it will interact with the entity.
 The relying party may choose to trust, not trust or partially trust the entity.
@@ -285,15 +285,15 @@ Like CWT and JWT, EAT uses COSE and JOSE to provide authenticity, integrity and 
 EAT places no new restrictions on cryptographic algorithms, retaining all the cryptographic flexibility of CWT, COSE, JWT and JOSE.
 
 EAT defines a means for nesting tokens and claims sets to accommodate composite devices that have multiple subsystems and multiple attesters.
-Full tokens with security envelopes may be embedded in an enclosing token.
+Tokens with security envelopes or bare claims sets may be embedded in an enclosing token.
 The nested token and the enclosing token do not have to use the same encoding (e.g., a CWT may be enclosed in a JWT).
 
-EAT adds the ability to detach claims sets and send them separately from a security enveloped EAT that contains a digest of the detached claims set.
+EAT adds the ability to detach claims sets and send them separately from a security-enveloped EAT that contains a digest of the detached claims set.
 
 This document registers no media or content types for the identification of the type of EAT, its serialization format or security envelope.
 The definition and registration of EAT media types is addressed in {{EAT.media-types}}.
 
-Finally, the notion of an EAT profile is introduced that facilitates the creation of narrowed definitions of EAT tokens for specific use cases in follow-on documents.
+Finally, the notion of an EAT profile is introduced that facilitates the creation of narrowed definitions of EATs for specific use cases in follow-on documents.
 
 
 ## Operating Model and RATS Architecture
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-eat.md
```diff
@@ -1208,21 +1214,22 @@ security state of the entity storing the private key used in a PoP application.
 
 # Detached EAT Bundles {#DEB}
 
-A detached EAT bundle is a structure to convey a fully-formed and signed token plus detached claims set that relate to that token.
-It is a top-level EAT message like a CWT or JWT.
-It can be occur any place that CWT or JWT messages occur.
-It may also be sent as a submodule.
+A detached EAT bundle is a structure to convey an EAT plus detached claims sets secured by that EAT.
+It is a top-level message like a CWT or JWT.
+It can be occur any place that a CWT or JWT occurs.
```
nedmsmith, line 1219:
```suggestion
It can occur any place that a CWT or JWT occurs.
```

draft-ietf-rats-eat.md
```diff
@@ -1208,21 +1214,22 @@ security state of the entity storing the private key used in a PoP application.
...
+It may be sent as a submodule.
```
nedmsmith, line 1220:
```suggestion
It may be contained within an EAT submodule.
```

draft-ietf-rats-eat.md
```diff
@@ -1208,21 +1214,22 @@ security state of the entity storing the private key used in a PoP application.
...
 
 A detached EAT bundle consists of two parts.
 
-The first part is a full top-level token.
-This top-level token MUST have at least one submodule that is a detached digest.
-This top-level token may be either CBOR or JSON-encoded.
+The first part is an encoded EAT.
```
nedmsmith, line 1224:
```suggestion
The first part is encoded as a traditional EAT.
```

draft-ietf-rats-eat.md
```diff
@@ -1208,21 +1214,22 @@ security state of the entity storing the private key used in a PoP application.
...
+This EAT MUST have at least one submodule that is a detached digest.
```
nedmsmith, line 1225:
```suggestion
This EAT MUST have at least one submodule that is a digest of a detached EAT bundle.
```

draft-ietf-rats-eat.md
```diff
@@ -1208,21 +1214,22 @@ security state of the entity storing the private key used in a PoP application.
...
+This EAT may be either CBOR or JSON-encoded.
```
nedmsmith, line 1226: Is isn't clear if the detatched claims set must be encoded using the same encoding as was used for the bounding EAT. Can the detatched EAT bundle be JSON encoded, while the EAT containing the digest of the bundle be CBOR encoded?

draft-ietf-rats-eat.md
```diff
@@ -1208,21 +1214,22 @@ security state of the entity storing the private key used in a PoP application.
...
 It MAY be a CWT, or JWT but MUST NOT be a detached EAT bundle.
```
nedmsmith, line 1227: Does 'it' refer to the EAT token or the detached EAT bundle?

draft-ietf-rats-eat.md
```diff
@@ -1208,21 +1214,22 @@ security state of the entity storing the private key used in a PoP application.
...
 It MAY also be some future-defined token type.
+It MUST have authenticity and integrity protection.
 The same mechanism for distinguishing the type for nested token submodules is employed here.
 
-The second part is a map/object containing the detached Claims-Sets corresponding to the detached digests in the full token.
+The second part is a map/object containing the detached Claims-Sets corresponding to the detached digests in the EAT.
```
nedmsmith, line 1232:
```suggestion
The second part is a map/object containing the detached Claims-Sets corresponding to their digests in the EAT.
```

draft-ietf-rats-eat.md
laurencelundblade: I think "is an encoded EAT" is quite accurate, but a little hard to get your head around.

It can be either JSON or CBOR-encoded. When CBOR is can be either a CWT or a UCCS and it is bstr wrapped. When JSON it is either a JWT or a UJCS. Note that UCCS and UJCS are not defined in this document. It could also be something else defined outside this document.

I tried using the term "full top-level token" to emphasize this, but I think "encoded EAT" is more accurate.

I don't like "traditional EAT" because the first part can be something not defined in the EAT document (e.g., UCCS or other future things (that I hope we don't defined)).

I don't want to change the wording here because I think it is precise and clean and the CDDL unfolds all the options here.

draft-ietf-rats-eat.md
laurencelundblade: Added a reference to detached submodule digest section.

draft-ietf-rats-eat.md
laurencelundblade: Added a comment that the bundle can be either CBOR or JSON which should clarify that mixing is allowed.

The mixing of encoding is very explicitly allowed here to accommodate composite devices where you don't control the serialization format of attesters in the components used to make up the composite device.

draft-ietf-rats-eat.md
laurencelundblade: See latest commit

PR #388, opened 2023-04-14
Clarify measurement result-id

laurencelundblade:
Text was a little vague...

This also renames "results-id" to "result-id" as it seems much more correct.

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -2480,6 +2480,7 @@ differences. A comprehensive history is available via the IETF Datatracker's rec
 - Rename Appendix E and clarify its purpose
 - Require presence of oemid claim if hwmodel is present; same for swversion and swname
 - Use normative language to describe the constrained device profile
+- Improve definitions in measurement results claim
 
 
 --- contributor
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -104,7 +104,7 @@ normative:
     target: https://portal.3gpp.org/desktopmodules/Specifications/SpecificationDetails.aspx?specificationId=729
     title: 3rd Generation Partnership Project; Technical Specification Group Core Network and Terminals; Numbering, addressing and identification
     author:
-    - org: 3GPP
+   - org: 3GPP
     date: 2019
 
   CoSWID: I-D.ietf-sacm-coswid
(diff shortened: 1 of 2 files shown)


quic-jodonogh (approved):

draft-ietf-rats-eat.md
```diff
@@ -972,8 +972,9 @@ For example, it may be desirable to report the results for measurements of the f
 Note that this claim is not for reporting the overall result of a verifier.
 It is solely for reporting the result of comparison to reference values.
 
-An individual measurement result is an array of two, an identifier of the measurement and an enumerated type that is the result.
-The range and values of the measurement identifier varies from one measurement scheme to another.
+An individual measurement result (individual-result) is an array of two, an identifier of the measurement (result-id) and an enumerated type that is the result (result).
```
nedmsmith, line 975:
```suggestion
An individual measurement result (individual-result) is an array consisting of two elements, an identifier of the measurement (result-id) and an enumerated type of the result (result).
```

draft-ietf-rats-eat.md
```diff
@@ -972,8 +972,9 @@ For example, it may be desirable to report the results for measurements of the f
...
+Different measurement systems will measure different things and perhaps measure the same thing in different ways.
```
nedmsmith, line 976:
```suggestion
Different measurement systems may measure different Target Environments, or perhaps measure the same Target Environment differently.
```

draft-ietf-rats-eat.md
```diff
@@ -972,8 +972,9 @@ For example, it may be desirable to report the results for measurements of the f
...
+It is up to each measurement system to define identifiers (result-id) for the measurements it reports.
```
nedmsmith, line 977:
```suggestion
It is up to each Attesting Environment to define the measurement identifiers (result-id) for the measurements it reports.
```

draft-ietf-rats-eat.md
laurencelundblade: While the target of an attestation is a target environment, the target of a measurement system here is not necessary a target environment. This is for use in many other contexts than measured boot or measurements in the TCG sense.

draft-ietf-rats-eat.md
laurencelundblade: Same comment as above. What is a measurement system is very open-ended here.

Issue #379, opened 2023-04-10
Security AD review:  clarify intuse claim


gmandyam:
Reference: https://mailarchive.ietf.org/arch/msg/rats/50ZbUkhSrU1cgOLYkir3f1kKFiY/
EAT reference:  https://www.ietf.org/archive/id/draft-ietf-rats-eat-19.html#name-intuse-intended-use-claim

"** Section 4.3.3.  intuse.  Does the WG expect this claim to be extended with additional values?  Should it be maintained with an IANA registry?"

gmandyam:
Extensibility should be clarified for this claim.  It seems that there could be desirable extensions that can be registered assuming they have been vetted through expert review.

PR #389, opened 2023-04-15
AD review:  clarify extensibility of "intuse" claim

gmandyam:
See https://github.com/ietf-rats-wg/eat/issues/379

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1174,7 +1174,7 @@ See {{common-types}} for OID and URI encoding.
 EATs may be used in the context of several different applications.  The "intuse"
 claim provides an indication to an EAT consumer about  the intended usage
 of the token. This claim can be used as a way for an application using EAT to internally distinguish between different ways it uses EAT.
-5 possible values for "intuse" are currently defined indicated 5 different intended use indications, but additional values above 5 can
+5 possible values for "intuse" are currently defined indicated 5 different intended use indications, but additional values can
 be defined and registered pending IANA-administered expert review (see {{iana-cons}}).  This could involve the future creation of a registry
 for additional values.
 
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1174,6 +1174,8 @@ See {{common-types}} for OID and URI encoding.
 EAT's may be used in the context of several different applications.  The "intuse"
 claim provides an indication to an EAT consumer about  the intended usage
 of the token. This claim can be used as a way for an application using EAT to internally distinguish between different ways it uses EAT.
+5 possible values for "intuse" are currently defined indicated 5 different intended use indications, but additional values above 5 can
+be defined and registered pending IANA-administered expert review (see {{iana-cons}}).
 
 1 -- Generic:
 : Generic attestation describes an application where the EAT consumer
@@ -1698,7 +1700,7 @@ an appropriate COSE hash algorithm, implying that an attacker cannot induce fals
 of a modified detached claims because the algorithms in the COSE registry are assumed to be
 of sufficient cryptographic strength.
 
-# IANA Considerations
+# IANA Considerations {#iana-cons}
 
 ## Reuse of CBOR and JSON Web Token (CWT and JWT) Claims Registries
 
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-eat.md
```diff
@@ -1174,6 +1174,8 @@ See {{common-types}} for OID and URI encoding.
 EAT's may be used in the context of several different applications.  The "intuse"
 claim provides an indication to an EAT consumer about  the intended usage
 of the token. This claim can be used as a way for an application using EAT to internally distinguish between different ways it uses EAT.
+5 possible values for "intuse" are currently defined indicated 5 different intended use indications, but additional values above 5 can
+be defined and registered pending IANA-administered expert review (see {{iana-cons}}).
```
carl-wallace, line 1178: EAT's should be EATs in the first sentence. I'd change the wording on the new language slightly:

```
Five possible values for "intuse" are currently defined, providing 5 different intended use indications. However, additional values can be defined and registered pending IANA-administered expert review (see {{iana-cons}}).
```

I don't think just referencing "expert review" answers the question regarding a registry, so I'd also add something about a possible future registry.

```
At present, there is not a registry for "intuse" values. If a future Internet Draft defines any additional values, a registry should be created and initialized with the five values defined in this document plus any new values.
```

gmandyam: Please review updates

draft-ietf-rats-eat.md
```diff
@@ -1171,11 +1171,12 @@ See {{common-types}} for OID and URI encoding.
 
 ### intuse (Intended Use) Claim
 
-EAT's may be used in the context of several different applications.  The "intuse"
+EATs may be used in the context of several different applications.  The "intuse"
 claim provides an indication to an EAT consumer about  the intended usage
 of the token. This claim can be used as a way for an application using EAT to internally distinguish between different ways it uses EAT.
 5 possible values for "intuse" are currently defined indicated 5 different intended use indications, but additional values above 5 can
```
quic-jodonogh, line 1177: "additional values above 5" reads slightly awkwardly to me. Maybe just "additional values"?

draft-ietf-rats-eat.md
carl-wallace: Works for me. Thanks.

draft-ietf-rats-eat.md
gmandyam: @quic-jodonogh change made.

laurencelundblade:
Agree that it should be possible to define new values.

I don't know if we can get away with saying that a registry can be created in the future, but am fine trying that.

Another thing possible here might be to say values above 64K are proprietary and below are reserved for a future IANA registry.

Agree with the wording suggestions.

carl-wallace:
There's a draft in LAMPS that has language on establishing a registry in the future (draft-ietf-lamps-header-protection).

draft-ietf-rats-eat.md
```diff
@@ -1171,9 +1171,12 @@ See {{common-types}} for OID and URI encoding.
 
 ### intuse (Intended Use) Claim
 
-EAT's may be used in the context of several different applications.  The "intuse"
+EATs may be used in the context of several different applications.  The "intuse"
 claim provides an indication to an EAT consumer about  the intended usage
 of the token. This claim can be used as a way for an application using EAT to internally distinguish between different ways it uses EAT.
```
laurencelundblade, line 1176: Wording could improve here with less use of "use".

draft-ietf-rats-eat.md
```diff
@@ -1171,9 +1171,12 @@ See {{common-types}} for OID and URI encoding.
...
+5 possible values for "intuse" are currently defined indicated 5 different intended use indications, but additional values can
+be defined and registered pending IANA-administered expert review (see {{iana-cons}}).  This could involve the future creation of a registry
+for additional values.
```
laurencelundblade, line 1179: Could improve the wording here and shorten a lot I think. Something like An IANA registry could be created to extend past the 5 values defined here."

laurencelundblade (changes requested):
Agree in concept, but please make a few wording improvements.

draft-ietf-rats-eat.md
gmandyam: Please see latest changes

laurencelundblade (approved):

Issue #377, opened 2023-04-10
Security AD Review:  clarify sec cons discussion on nonces


gmandyam:
Reference:  https://mailarchive.ietf.org/arch/msg/rats/50ZbUkhSrU1cgOLYkir3f1kKFiY/

EAT reference:  https://www.ietf.org/archive/id/draft-ietf-rats-eat-19.html#name-multiple-eat-consumers

"** Section 9.4.

   However, assume the EAT of the previous example is hierarchical and
   each claim subset for a downstream consumer is created in the form of
   a nested EAT.  Then, Transport Layer Security between the receiving
   and downstream consumers is not strictly required.  Nevertheless,
   downstream consumers of a nested EAT should provide a nonce unique to
   the EAT they are consuming.

I don't follow how a hierarchical relation changes the required security.  Could this be clarified?"

PR #390, opened 2023-04-15
AD review:  clarify sec cons discussion on nonces

gmandyam:
See https://github.com/ietf-rats-wg/eat/issues/377

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -1683,8 +1683,10 @@ downstream consumer should leverage a communication security protocol
 
 However, assume the EAT of the previous example is hierarchical and
 each claim subset for a downstream consumer is created in the form of
-a nested EAT.  Then the nested EAT is itself cryptographically-verifiable
-by a downstream consumer.  Therefore, Transport Layer Security between the receiving and
+a nested EAT.  Then the nested EAT is itself cryptographically-verifiable (due to its
+COSE envelope)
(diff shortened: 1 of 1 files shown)


carl-wallace (approved):

quic-jodonogh (approved):

quic-jodonogh (approved):

draft-ietf-rats-eat.md
```diff
@@ -1683,7 +1683,10 @@ downstream consumer should leverage a communication security protocol
 
 However, assume the EAT of the previous example is hierarchical and
 each claim subset for a downstream consumer is created in the form of
-a nested EAT.  Then, Transport Layer Security between the receiving and
+a nested EAT.  Then the nested EAT is itself cryptographically-verifiable (due to its
```
laurencelundblade, line 1686: Don't think cryptographically-verifiable should be hyphenated.

draft-ietf-rats-eat.md
```diff
@@ -1683,7 +1683,10 @@ downstream consumer should leverage a communication security protocol
...
+COSE envelope)
```
laurencelundblade, line 1687: Maybe mention COSE encrypted too since that's in the above example and because the role of TLS here is more about privacy than authentication/integrity.

laurencelundblade (changes requested):

laurencelundblade (approved):

PR #391, opened 2023-04-15
Rework entity definition to align with RAT architecture

laurencelundblade:


gmandyam:
Do we need to say anything at all about layered attestation as defined in RFC 9334?

For instance, with layered attestation the following requirement appears in the definition in RFC 9334:  "The bottom layer of an Attester has an Attesting Environment that is typically designed to be immutable or difficult to modify by malicious code."  It is conceivable that a mutable attester that may be vulnerable to runtime modification by an attacker  (e.g. Android SafetyNet) could provide a EAT token with nesting.  It would not be an ideal implementation of EAT, but depending on the application it may be considered sufficient.  Tying EAT nesting to the 9334 definition of layered attestation may restrict intended applications of EAT.

draft-ietf-rats-eat.md
```diff
@@ -237,9 +237,13 @@ EAT is built on CBOR Web Token (CWT) {{RFC8392}} and JSON Web Token (JWT) {{RFC7
...
 
-The document uses the term "entity" to refer to the target of an EAT. Many of the claims defined in this document are claims about an entity, which is equivalent to an attesting environment as defined in [RATS.architecture]. An entity may be the whole device, a subsystem, a subsystem of a subsystem, etc.
-Correspondingly, the EAT format allows claims to be organized using mechanisms like submodules and nested EATs (see {{submods}}).
-The entity to which a claim applies is the submodule in which it appears, or to the top-level entity if it doesn't appear in a submodule.
+The document uses the term "entity" to refer to the target of an EAT.
+Most of the claims defined in this document are claims about an entity.
+An entity is equivalent to an attester target environment as defined in [RATS.architecture].
+
+Layered attestation and composite devices as described in [RATS.architecture] are supported by a submodule mechanism (see {{submods}}).
+This mechanism allows one EAT to have several submodule EATs and for multiple layers of nesting.
```
quic-jodonogh, line 245: This reads a bit awkwardly to me. Something like "This mechanism allows one EAT to have several submodule EATs and supports multiple layers of nesting".

laurencelundblade:
I know layered attestation in 9334 is not the best description, but I think the text I proposed works OK and doesn't cause any problems. We'll definitely sail through better if we align as much as we can with 9334 and particularly if we don't make any fuss about it.

Let me know if you think the text I propose has any problems.

gmandyam (approved):

PR #393, opened 2023-04-17
Move verification key origin paragraph from appendix to sec cons

laurencelundblade:


draft-ietf-rats-eat.md
```diff
@@ -1698,6 +1698,19 @@ an appropriate COSE hash algorithm, implying that an attacker cannot induce fals
 of a modified detached claims because the algorithms in the COSE registry are assumed to be
 of sufficient cryptographic strength.
 
+## Verification Keys {#verfication-key-sc}
+
+In all cases there must be some way that the verification key is itself verified or determined to be trustworthy.
+The key identification itself is never enough.
+This will always be by some out-of-band mechanism that is not described here.
+For example, the verifier may be configured with a root certificate or a master key by the verifier system administrator.
```
nedmsmith, line 1706:
```suggestion
For example, the verifier may be configured with a root certificate or a master key by a trusted entity such as the verifier system administrator.
```

draft-ietf-rats-eat.md
```diff
@@ -1698,6 +1698,19 @@ an appropriate COSE hash algorithm, implying that an attacker cannot induce fals
...
+
+Often an X.509 certificate or an endorsement carries more than just the verification key.
```
nedmsmith, line 1708: An X.509 certificate could be regarded as an Endorsement conceptual message.

nedmsmith, line 1708:
```suggestion
Often an Endorsement, such as an X.509 certificate, carries more than just the verification key.
```

draft-ietf-rats-eat.md
```diff
@@ -1698,6 +1698,19 @@ an appropriate COSE hash algorithm, implying that an attacker cannot induce fals
...
+For example, an X.509 certificate might have key usage constraints and an endorsement might have reference values.
```
nedmsmith, line 1709:
```suggestion
For example, an X.509 certificate might have key usage constraints and an Endorsement might have Reference Values.
```

draft-ietf-rats-eat.md
```diff
@@ -1698,6 +1698,19 @@ an appropriate COSE hash algorithm, implying that an attacker cannot induce fals
...
+When this is the case, the key identifier must be either a protected header or in the payload such that it is cryptographically bound to the EAT.
```
nedmsmith, line 1710:
```suggestion
When this is the case, the key identifier must be either a protected header or in the payload, such that it is cryptographically bound to the EAT.
```

draft-ietf-rats-eat.md
carl-wallace: The suggestion to capitalize these terms is contrary to comments received and acted upon last summer (from Eliot I think). The change took effect in -15.

draft-ietf-rats-eat.md
laurencelundblade: The word "trusted" is intentionally never used in the EAT document because to me it is too subjective. Trusted for what purpose and against what level of attack?

draft-ietf-rats-eat.md
laurencelundblade: Prefer to leave it open since we have only a conceptual definition of endorsement.

draft-ietf-rats-eat.md
laurencelundblade: Yes, we're using lower case.

PR #394, opened 2023-06-12
Clarification on new UEID types

laurencelundblade:


gmandyam:
>The different types of UEIDs accommodate different manufacturing processes, accommodate small UEIDs, provide an option that doesn't require registration fees and central administration.

Please add back 1), 2), 3).

PR #398, opened 2023-06-26
Clarity that DLOAs must be authentic

laurencelundblade:


quic-jodonogh (approved):
A very minor nit - possibly not worth changing the PR for:
"Clarity that DLOAs must be authentic" -> "Clarify that..."
I've approved anyway as this is such a small thing.

gmandyam (approved):

PR #402, opened 2023-08-10
Fixes requested by IANA

laurencelundblade:


carl-wallace:
This looks fine but you missed one item from the review. 

```
IANA QUESTION -> Should the Claim Name be changed to "OEM Authorized Boot" instead, or is "OEM Authortised Boot" correct? Also, should the description be changed to "Indicates whether the software booted was OEM authorized"?
```

carl-wallace (approved):

PR #404, opened 2023-09-04
Add section on partial and full profiles

laurencelundblade:


draft-ietf-rats-eat.md
```diff
@@ -1287,6 +1287,31 @@ A profile document doesn't have to be in any particular format. It may be simple
 
 A profile may define, and possibly register, one or more new claims if needed. A profile may also reuse one or more already defined claims, either as-is or with values constrained to a subset or subrange.
 
+
+## Full and Partial Profiles
+
+A "full" profile is one that fully guarantees interoperability when a sender and receiver both adhere to it.
```
carl-wallace, line 1293: Should a profile include trust anchor key material or a reference to key material? This is required to "fully guarantee  interoperability". At present, only a key ID is included.

draft-ietf-rats-eat.md
```diff
@@ -1287,6 +1287,31 @@ A profile document doesn't have to be in any particular format. It may be simple
...
+
+Partial profiles are discouraged.
+It is better to define several full profiles perhaps derived from each other.
+That way each profile is actually usable.
+It is relatively easy and inexpensive to define profiles as they don't have to be standards track and don't have to be registered anywhere.
+For example, rather than leaving the signing algorithms unspecified perhaps to accommodate post-quantum algorithms, a profile can specify the NIST signing algorithms.
+When the post-quantum algorithms are selected, a derived profile that adds the new algorithms can be specified.
+The derived profiles can be as simple as stating "the 'Xxx PQ' profile is the same as the 'Xxx' profile with the addition that the receiver must implement the 'yyy' post-quantum algorithm."
+
+A "eat_profile" claim SHOULD NOT be used to identify partial profiles.
```
carl-wallace, line 1312: Should this be MUST NOT? When would it be OK to use eat_profile to identify a partial profile?

draft-ietf-rats-eat.md
laurencelundblade: #417 

It seems like a slippery slope to start discussing key formats, key distribution and key trust.

carl-wallace: There's no need to get into discussing key formats, etc. Maybe something like:

```
A "full" profile is one that fully guarantees interoperability when a sender and receiver both adhere to it and have access to any necessary verification, decryption or MAC key(s), as discussed in sections 6.2.9 and 9.6.
```

gmandyam:
Lack of a specific alg. identifier does not imply that interoperability is not possible.  For instance, https://datatracker.ietf.org/doc/html/rfc8152#appendix-A.1 discusses the use of implicit algorithms.  In my interpretation of the PQC example, if the alg. field is left empty but the kid points to a PQC keypair then interoperability is still possible without having to re-define the profile to explicitly include the PQC algm.

carl-wallace:
Re: specific algorithm and interoperability being possible, my point was that "fully guaranteed" doesn't fit because the recipient might have no idea where to get the verification key material. The "fully guaranteed" verbiage is new, and doesn't strike me as quite right without considering that. Likewise for decryption key.

laurencelundblade:
> Lack of a specific alg. identifier does not imply that interoperability is not possible. For instance, https://datatracker.ietf.org/doc/html/rfc8152#appendix-A.1 discusses the use of implicit algorithms. In my interpretation of the PQC example, if the alg. field is left empty but the kid points to a PQC keypair then interoperability is still possible without having to re-define the profile to explicitly include the PQC algm.

Yes, this is true and with the it's OK to to not have an alg ID in a full profile.

laurencelundblade:
Please take a look at my recent update here. It tightens up a lot and I think makes the notion of a full profile much more clear. Also this hopefully addresses Simon's comment too.

draft-ietf-rats-eat.md
```diff
@@ -1295,6 +1295,29 @@ A profile document doesn't have to be in any particular format. It may be simple
...
+Profile specifications may assume the receiver has access the necessary verification keys or may go into detail on the means to access verification keys.
+
+The "eat_profile" claim MUST NOT be used to identify partial profiles.
+
+While fewer profiles are preferrable, sometimes several may be needed for a use case.
+One approach to handling variation in devices might be to define several full profiles that are variants of each other.
+It is relatively easy and inexpensive to define profiles as they don't have to be standards track and don't have to be registered anywhere.
+For example, flexibility for post-quantum algorithms can be handled as follows.
+First, define a full profile for a set of non-post-quantum algorithms for current use.
+Then, when post-quantum algorithms are settled, define another full profile derived from the first.
```
thomas-fossati, line 1318: _OOP is back on track!_ 😄 

Just a clarification: from this example I understand that "deriving" allows overriding what is inherited?

An alternative would be to define a partial profile as an abstract class with a "pure virtual" signing algorithm that full profiles must instantiate.

laurencelundblade: Hi Thomas,

This paragraph is just an example, not normative so you can do what you want.

So, yes, a derived profile can override another or you can do an abstract class profile.

Personally, I have a preference to avoid partial profiles because I prefer to go directly from the abstraction that EAT is to a fully specified protocol in one step because I think it is less confusing for the implementors, but no objection for folks that want them.

SimonFrost-Arm (approved):
LGTM

PR #406, opened 2023-09-07
clarify OEMID lifetimes

laurencelundblade:


draft-ietf-rats-eat.md
```diff
@@ -600,11 +600,18 @@ The "oemid" claim identifies the Original Equipment Manufacturer (OEM) of the ha
 Any of the three forms described below MAY be used at the convenience of the claim sender.
 The receiver of this claim MUST be able to handle all three forms.
 
+Note that the "hwmodel" claim in {{hwmodel}}, the "oemboot" claim in {{oemboot}} and "dbgstat" claim in {{dbgstat}} depend on this claim.
```
carl-wallace, line 603: Should there be a similar requirement for oemid to be present when dbgstat or hwmodel are present as #407 added for oemboot?

laurencelundblade: There is a requirement for hwmodel and hwversion. Added one for dbgstat.

PR #415, opened 2023-09-09
Why manifest sigs should be retained

laurencelundblade:


draft-ietf-rats-eat.md
```diff
@@ -905,6 +905,7 @@ The purpose of these claims in an EAT is to relay them without modification to t
 Some manifests are signed by their software manufacturer independently, and some are not either because they do not support signing or the manufacturer chose not to sign them.
 For example, a CoSWID might be signed independently before it is included in an EAT.
 When signed manifests are put into an EAT, the manufacturer's signature SHOULD be included even though an EAT's signature will also cover the manifest.
+This allows the receiver to verify the manifest directly with the manufacturer if they wish.
```
carl-wallace, line 908: The phrase "directly with" connotes some interaction with the manufacturer that isn't necessary. Maybe "This allows the receiver to verify the manufacturer originated the manifest"

laurencelundblade: Thx Carl. Think I fixed. Let me know what you think.

PR #417, opened 2023-09-12
clarify verification key in standard profile

laurencelundblade:


carl-wallace (commented):
I think you mean:
```
 (It is assumed the receiver has access to a database of trusted verification keys which allows lookup of a key given a verification key ID; the key format and means of distribution are beyond the scope of this profile) 
```
But, I think I'd address this not here but where "fully guaranteed" is introduced.

laurencelundblade:
I have this text and also text in the full/partial profile PR. I like having both because I think it gives more clarity on something that is confusing. Will keep them both unless there is strong objection.

Thanks as always for the reviews.

SimonFrost-Arm (approved):
LGTM

PR #418, opened 2023-09-15
minor typo fix in the "constrained profile" table

thomas-fossati:
Also, while we are in the area, there's an opportunity to shorten the profile identifier using [RFC2648](https://datatracker.ietf.org/doc/rfc2648/) URNs, i.e.: `urn:ietf:id:ietf-rats-eat-21` and `urn:ietf:rfc:XXXX` on publication.


carl-wallace:
This looks like a duplicate of #411

thomas-fossati:
Oops, yes, I hadn't seen it, sorry!

Issue #419, opened 2023-09-15
use rfc2648 for profile identification


thomas-fossati:
_(This is just a suggestion that I leave to the discretion of the editors.)_

There's an opportunity to shorten the Constrained Device Standard Profile
identifier using [RFC2648](https://datatracker.ietf.org/doc/rfc2648/) URNs, i.e., using `urn:ietf:rfc:XXXX` rather than `https://www.rfc-editor.org/rfc/rfcXXXX`.

This would remove 21 bytes from the wire per token, which is in line with the "constrained device" scope we are trying to cover here.

laurencelundblade:
Fixed in #420

PR #420, opened 2023-09-18
Use urn:ietf for profile ID

laurencelundblade:
Address #419 

This saves a lot of bytes and is prettier


thomas-fossati (approved):
👍

SimonFrost-Arm (approved):
+1

PR #422, opened 2023-09-24
Correct the JSON DEB example

laurencelundblade:
The JSON DEB example is now a correct JWT with correct content that can be validated.

Also checked in the JSON files used to construct the example.

There was a "TODO" tag for this that no noticed. Searching for "TODO" now has no results.


thomas-fossati (approved):
👍

thomas-fossati:
> The nonces

s/nonces/ueids/


> in the example are Type 1 but are only 112 bits. The prose for type 1 requires 128, 192 or 256. The values in the examples are:
> 
> ```
> 01 98 F5 0A 40 36 C0 58 61 C8 86 0D 13 A6 38
> ```
> 
>  ```
> 01 BE D4 02 2E F2 7D F0 7A EB 6D FC B3 3C F3
> ```

yep, these need a couple more bytes

PR #423, opened 2023-09-25
Fix UEID length and add simple example to intro

laurencelundblade:
A 48-bit mac address formats to 7 binary bytes. b64url encoding of 7 bytes is 10 bytes, not 12. The previous mistake was to include the padding for b64 encoding which is not present with b64 encoding.

Warren Kumari suggested a simple example be added early in the document as part of the larger review. Seems like a good idea! Make it easier for people to understand that aren't very familiar with attestation, which is a lot of people since attestation is not widely used yet.

### DIFF
--- a/cddl/ueid.cddl
+++ b/cddl/ueid.cddl
@@ -1,3 +1,3 @@
 $$Claims-Set-Claims //= (ueid-label => ueid-type)
 
-ueid-type = JC<base64-url-text .size (12..44) , bstr .size (7..33)>
+ueid-type = JC<base64-url-text .size (10..44) , bstr .size (7..33)>
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -225,6 +225,23 @@ An EAT may be encoded in either JavaScript Object Notation (JSON) {{RFC8259}} or
 EAT is built on CBOR Web Token (CWT) {{RFC8392}} and JSON Web Token (JWT) {{RFC7519}} and inherits all their characteristics and their security mechanisms.
 Like CWT and JWT, EAT does not imply any message flow.
 
+Following is a very simple example.
+It is JSON format for easy reading, but could also be CBOR.
+Only the Claims-Set, the payload for the JWT, is shown.
+
+~~~~
+{::include cddl/Example-Payloads/simple.json}
+~~~~
+
+This example has a nonce for freshness.
+This nonce is the base64 encoding of a 12 byte random binary byte string.
+The ueid is effectively a serial number uniquely identifying the device.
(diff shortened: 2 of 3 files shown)


draft-ietf-rats-eat.md
```diff
@@ -225,6 +225,23 @@ An EAT may be encoded in either JavaScript Object Notation (JSON) {{RFC8259}} or
...
+~~~~
+{::include cddl/Example-Payloads/simple.json}
+~~~~
+
+This example has a nonce for freshness.
+This nonce is the base64 encoding of a 12 byte random binary byte string.
+The ueid is effectively a serial number uniquely identifying the device.
+This ueid is the base64 encoding of a 48-bit MAC address.
+The oemid identifies the manufacturer using a Private Enterprise Number {{PEN}}.
+The SW is identified by a simple string name and version.
```
thomas-fossati, line 241:
```suggestion
The software is identified by a simple string name and version.
```

thomas-fossati (approved):

draft-ietf-rats-eat.md
```diff
@@ -225,6 +225,23 @@ An EAT may be encoded in either JavaScript Object Notation (JSON) {{RFC8259}} or
...
+Only the Claims-Set, the payload for the JWT, is shown.
+
+~~~~
+{::include cddl/Example-Payloads/simple.json}
+~~~~
+
+This example has a nonce for freshness.
+This nonce is the base64 encoding of a 12 byte random binary byte string.
+The ueid is effectively a serial number uniquely identifying the device.
+This ueid is the base64 encoding of a 48-bit MAC address.
```
carl-wallace, line 239: PR looks good. One minor comment. It may be worth calling out the type byte. Perhaps: "This ueid is the base64 encoding of a 48-bit MAC address along with a type byte set to 0x02."

thomas-fossati: maybe it's also s good case for using the term `base64url-encoded` from the document glossary, as in:
```
This nonce is the base64url-encoded representation of a 12 byte random binary byte string.
```

cddl/ueid.cddl
```diff
@@ -1,3 +1,3 @@
 $$Claims-Set-Claims //= (ueid-label => ueid-type)
 
-ueid-type = JC<base64-url-text .size (12..44) , bstr .size (7..33)>
+ueid-type = JC<base64-url-text .size (10..44) , bstr .size (7..33)>
```
thomas-fossati, line 3: good catch!

laurencelundblade:
Fixes suggested by Carl and Thomas for b64 and ueid.

PR #424, opened 2023-09-26
Constrain the content of a nested JSON token.

quic-jodonogh:
This is how I'd propose to constrain nested JSON to be a valid EAT token.

I added a definition of the .json operator as this is currently only defined in an early-stage I-D.


quic-jodonogh:
We have agreed not to do the JSON operator thing. I'll write a separate PR for the easy fixes.

PR #429, opened 2023-10-07
Clarify varying MAC addresses can't be used as UEID

laurencelundblade:


carl-wallace (changes requested):
The "some devices" part does not seem right. Suggest:

```
MAC addreses that not permanent in accordance with requirement in this document MUST not be used for the UEID
```

PR #433, opened 2023-12-09
Minor OID use clarification

laurencelundblade:
The RFC 9165 reference is needed for the CDDL .feature control operator.


nedmsmith (approved):
All these changes are improvements IMO.

PR #439, opened 2024-01-06
Types in CWT registry are all CDDL definitions

laurencelundblade:
See https://mailarchive.ietf.org/arch/msg/cbor/qMs19DfPFSOrQ-e0tRHk-3Zaer4/

The "Claim Value Type(s)" column is now all CDDL definitions.

Some of the types were in error and corrected.

This only affected the IANA section and only parts of it that aren't really normative.


laurencelundblade:
FAIL with GitHub and other stuff. Will try again.

laurencelundblade:
Still messed up...

PR #443, opened 2024-01-14
Fix two "byte string" occurrences in IANA section

laurencelundblade:


laurencelundblade:
Github problems; PR didn't come out right

PR #450, opened 2024-03-31
Fix/update examples and CDDL validation

laurencelundblade:
This updates the examples with the IANA-assigned labels and tags.

This fixes some errors in the examples, mostly missing commas.

There is a change to the normative CDDL to compensate for a bug in the cddl tool. RFC 8610 says that there is a problem with generics nesting in the cddl tool. This is presumed to be the cause of the problem. The solution is to provide a non-nested version of the JC<> generic for use in contexts where the cddl validation gets confused. (It didn't produce an error, it just didn't validate correct input; the solution was found by extensive trial and error).


cabo:
Can we fix this now?

I believe the putative workaround for an undocumented problem and the explanation

```
; Same as JC<> but with unwound generic nesting as it seems to cause
; problems. Perhaps this is the nesting problem described in RFC
; 8610.
JC-NEST-SAFE<J,C> = J .feature "json" / C .feature "cbor"
```

demonstrates an incredible level of incompetence in handling tool bugs; this should not be in an RFC.

PR #456, opened 2024-05-23
Address Roman's May 2024 comments

laurencelundblade:
I believe the idnits complaint about  "/    the    name" is about comments in the CBOR diagnostic notation in the examples.

- Reference to RATS Architecture, RFC 9334, is made informative since the document is informational.

- Reference to Long Lines in Draft, RFC 8792, is made informative since the document is informational.

- Reworded full profiles description to convey intent without using "may not"

- Upated references for UUIDs and LDAP to non-obsolete documents.

- Removed some non-ascii quote marks

- "MAY not" -> "MAY NOT"

### DIFF
--- a/draft-ietf-rats-eat.md
+++ b/draft-ietf-rats-eat.md
@@ -74,15 +74,13 @@ normative:
   RFC8259:
   RFC8392:
   RFC8610: cddl
-  RFC8792:
   RFC3986:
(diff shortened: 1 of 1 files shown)


gmandyam (changes requested):
Return RFC 9334 to normative reference

PR #459, opened 2024-07-26
Update intuse registry text

gmandyam:


draft-ietf-rats-eat.md
```diff
@@ -2052,7 +2053,7 @@ Description:
 : This is a text paragraph or more that sufficiently defines what the intended use means. It may also be a reference to another document.
 
 
-These 5 values are the initial content of the registry:
+These 5 values are the initial content of the registry are listed below.  Note that 0 will not be assigned for the CBOR value, and the maximum CBOR value for assignment is 255.
```
thomas-fossati, line 2056:
```suggestion
The following 5 values represent the initial content of the registry.  Note that 0 will not be assigned for the CBOR value, and the maximum CBOR value for assignment is 255.  
```

ababer:
IANA has a strong preference for not including the word "Registries" (or, although it was once common, "Parameters") in the name of a registry or group of registries, unless it would be confusing in that particular case to leave it out. The reasoning is that all registries and groups could include the word "Registry" or "Registries" in the title, but most of them don't. 

Would you want to use "Remote Attestation Procedures (RATS)" for the registry group name? That would get the acronym "RATS" into the list at https://www.iana.org/protocols, which could be useful for find-in-page searchability.

One other note: in the next line, "Expert Review Required" should be changed to "Expert Review" (the RFC 8126 term).

draft-ietf-rats-eat.md
```diff
@@ -2029,10 +2029,11 @@ specification reference.
...
 
-IANA is requested to create a new registry titled "EAT Intended Uses".
+IANA is requested to create a new registry titled "EAT Intended Uses" in a new registry group 
+called "Remote Attestation Procedures (RATS)."
 The registry uses the "Expert Review Required" registration procedure {{RFC8126}}.
 
-Some guidelines for experts:
+Guidelines for experts:
 
 * Each intended use should be clearly described so a user of can know what it means.
```
laurencelundblade, line 2038: typo (mine)

Need to add "it"... "as user of it can know"

laurencelundblade (approved):

ababer:
Just one IANA request left: please change "Expert Review Required" to "Expert Review."

gmandyam:
> Just one IANA request left: please change "Expert Review Required" to "Expert Review."
> 
> One additional question: should "EAT Intended Uses" be changed to "Entity Attestation Token (EAT) Intended Uses"? (I don't have a recommendation here.)

Added clarifications.  Please review.

PR #463, opened 2024-08-11
Fix editorial issues

steven-bellock:
- Expand contractions.
- Remove double spaces.
- Remove redundant "always" from "MUST always".
- "MUST never" to "MUST NOT".
- Various spelling and grammar fixes.

Signed-off-by: Steven Bellock <sbellock@nvidia.com>


gmandyam (approved):
Looked through proposed changes and they are all acceptable.

laurencelundblade (approved):
Thanks very much!

carl-wallace (approved):

PR #468, opened 2024-10-09
Rename EAT-CBOR-Tagged-Token to CBOR-Tagged-Token

laurencelundblade:
This is renamed because the tagged token contained might not be EAT format. For example it could be from draft-ietf-rats-msg-wrap.

There is no need to change anything else to accommodate draft-ietf-rats-msg-wrap. The socket name was just misleading.


thomas-fossati (approved):
thanks!

PR #469, opened 2025-01-17
Change from NaN to NULL for heading in location claim

laurencelundblade:
This changes the heading value from NaN to NULL to represent the heading when stationary.

This aligns with a change in the W3C specification.

Also, JSON can't represent a NaN and the convention in JSON is to use NULL.

(NaN is IEEE 754 representation of "not a number" typically used to represent divide by 0 in local calculations).

This is somewhat of a protocol change, but
- The protocol was unimplementable in JSON
- Backwards compatibility in CBOR is easy to provide by checking for both NaN and NULL


draft-ietf-rats-eat.md
```diff
@@ -834,7 +834,7 @@ This level indicates that all debug facilities for the entity are permanently di
 The "location" claim gives the geographic position of the entity from which the attestation originates.
 Latitude, longitude, altitude, accuracy, altitude-accuracy, heading and speed MUST be as defined in the W3C Geolocation API {{W3C.GeoLoc}}
 (which, in turn, is based on {{WGS84}}).
-If the entity is stationary, the heading is NaN (floating-point not-a-number).
+If the entity is stationary, the heading is NULL.
```
thomas-fossati, line 837: There seems to be no reason for the capitalisation.
(We could use the backticks to show it’s from the CDDL prelude.)

```suggestion
If the entity is stationary, the heading is `null`.
```

PR #470, opened 2025-01-18
Fix errors in detached eat bundle example

laurencelundblade:


gmandyam:
Original hex for the TEE submod (266 index) had upper case hex.  Why was it changed to lower case?

PR #474, opened 2025-04-09
Better description of unsecured JWT

laurencelundblade:
Use of the term NULL or null is not really right.


draft-ietf-rats-eat.md
```diff
@@ -1353,7 +1353,8 @@ Often a tag is unnecessary because the surrounding or carrying protocol identifi
 ### COSE/JOSE Protection {#message-type}
 
 COSE and JOSE have several options for signed, MACed and encrypted messages.
-JWT may use the JOSE NULL protection option.
+It may be an Unsecured JWT as described in Section 6 of {{RFC7519}}.
```
cabo, line 1356: +1.

Suggestion:

```suggestion
It may be an Unsecured JWT as described in {{Section 6 of RFC7519}}.
```

cabo (approved):

